PR #54, opened 2025-08-28
Add a patch to fix the generating spinner freezing when CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC is set

heromantf:
…ISABLE_NONESSENTIAL_TRAFFIC is set


bl-ue:
Thank you!

PR #139, opened 2025-10-31
Add fnm multishell path to the search dirs

wu-json:
FNM users may have claude paths that resolve to something like `/Users/me/.local/state/fnm_multishells/83926_1761897468974/bin/claude`.


bl-ue:
Thank you!

PR #162, opened 2025-11-06
Implement native support

### LINKED ISSUE #29: Support native CC installations
Installing Claude Code with `winget` downloads an EXE which tweakcc cannot modify.  It might be possible to open the executable and modify the JS that's likely contained in it, but it will be harder.  There may also be fixed width sections or hashes in the executable that we would need to somehow when we write the JS code.

### LINKED ISSUE #131: In my claude its ~/.local/bin/claude -> /root/.local/share/claude/versions/2.0.28
That's where my symlink goes after a claude install on Ubuntu 24.04

Currently running: native (2.0.28)
 └ Path: /root/.local/bin/claude
 └ Invoked: /root/.local/share/claude/versions/2.0.28
 └ Config install method: native
 └ Auto-updates enabled: default (true)
 └ Search: OK (bundled)

so ... not sure where cli.js would be ?!

signadou:
Closes #29
Closes #131

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- Support slight bun-specific differences in React module accessing, `$` identifier frequency, and unicode escaping (#163) - @bl-ue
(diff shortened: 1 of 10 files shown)


ghost:
This is a massive achievement for all of Clawd-kind!

PR #183, opened 2025-11-11
Fix Mach-O segment alignment for ARM64 binaries

### LINKED ISSUE #180: with homebrew installed native CC, repackMachO corrupts __LINKEDIT segment causing SIGKILL
Tried to debug my problem with Claude, so not 100% confident landed on the right diagnosis.

After tweakcc apply on Claude Code 2.0.36 (native binary via Homebrew), the binary becomes malformed and immediately gets SIGKILL.

Issue:
When tweakcc extends the __BUN segment by 24KB, it corrupts the __LINKEDIT segment:

__LINKEDIT filesize decreased from 1462352 to 1461488 (-864 bytes)
This should not decrease when extending other segments
Results in "Malformed Mach-o file" error from dtrace
Binary gets SIGKILL immediately on execution

Diff of otool output showing the corruption:
__BUN segment:
- vmsize: 0x5fd0000 → 0x5fd6000 (+24KB)
- filesize: 100466688 → 100491264 (+24KB)

__LINKEDIT segment:
- vmaddr: 0x9d64000 → 0x9d6a000 (+24KB) ✓ correct
- fileoff: 160350208 → 160374784 (+24KB) ✓ correct  
- filesize: 1462352 → 1461488 (-864 bytes) ✗ WRONG - corrupts linking data

Expected: __LINKEDIT filesize should remain unchanged or be recalculated properly, not decreased.

signadou:
Closes #180

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- Fix Mach-O segment alignment for ARM64 binaries (#183) - @signadou
+
 ## [v3.0.0](https://github.com/Piebald-AI/tweakcc/releases/tag/v3.0.0) - 2025-11-10
 
 - Remove the "CLAUDE CODE" figlet ASCII art customization to keep up with Claude Code (#174) - @bl-ue


src/utils/nativeInstallation.ts
```diff
@@ -781,12 +781,19 @@ function repackMachO(
     const sizeDiff = newSectionData.length - Number(bunSection.size);
 
     if (sizeDiff > 0) {
-      // CRITICAL: Round up to page alignment (4096 bytes)
+      // CRITICAL: Round up to page alignment
```
bl-ue, line 784:
```suggestion
      // CRITICAL: Round up to page alignment
      // See #180.
```

PR #193, opened 2025-11-12
Improve POSIX fallback for Claude Code detection #191

### LINKED ISSUE #191: Try to find cli.js by the equivalent of `readlink $(which claude)` when all else fails
As suggested by @Mehdi-Bl in #189.

Mehdi-Bl:
As discussed in Try to find cli.js by the equivalent of readlink $(which claude) when all else fails #191

Closes #191

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -8,6 +8,7 @@ import {
   CLIJS_SEARCH_PATH_INFO,
   CONFIG_FILE,
   CONFIG_DIR,
+  PATH_CHECK_TEXT,
 } from './utils/types.js';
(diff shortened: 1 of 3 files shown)


Mehdi-Bl:
I will fix and double check again.
Should have run linters too.

bl-ue:
Hey @Mehdi-Bl, are you still working this?  I'm happy to take it over if you're busy.

Mehdi-Bl:
Go ahead, will not be able before 2 days.

bl-ue:
Thank you!  I'm going to refactor this a bit more after I merge it to close some more issues.

PR #226, opened 2025-11-24
Emit a Rust-style error when system prompt markdown files contain unescaped backslashes

### LINKED ISSUE #205: Expected CommonJS module to have a function wrapper. If you weren't messing around with Bun's internals, this is a bug in Bun
Apply breaks Claude 2.0.42, even without any specific customization.

claude
 6 |     "use strict";
 7 |
 8 |     //
 9 |     if (entry.instantiate)
10 |         return entry.instantiate;
11 |     var instantiatePromise = (async () => {
                                               ^
TypeError: Expected CommonJS module to have a function wrapper. If you weren't messing around with Bun's internals, this is a bug in Bun

bl-ue:
Closes #205 (again)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-- Emit a Rust-style error when system prompt markdown files contain unescaped backslashes (#226) - @bl-ue
+- Emit a Rust-style error when system prompt markdown files contain unescaped backslashes (#226) - @bl-ue & @mike1858
 
 ## [v3.1.2](https://github.com/Piebald-AI/tweakcc/releases/tag/v3.1.2) - 2025-11-23
 


src/utils/promptSync.ts
```diff
@@ -1045,6 +1045,196 @@ const buildSearchRegexFromPieces = (
...
+        results.get(line)!.push(column);
+      }
+    }
+
+    i++;
+    column++;
+  }
+
+  return results;
+};
```
mike1858:
```suggestion
export const findUnescapedBackticks = (content: string) => {
  const result = new Map<number, number[]>();
  let depth = 0, line = 1, col = 1; // depth inside ${...}, 1-based line/column

  for (let i = 0; i < content.length; i++, col++) {
    const ch = content[i];

    // Track newlines and reset column
    if (ch === '\n') { line++; col = 0; continue; }

    const next = content[i + 1];

    // Enter interpolation on ${
    if (ch === '$' && next === '{') { depth++; i++; col++; continue; }

    // While inside interpolation, track nested { } depth and skip backtick checks
    if (depth) {
      if (ch === '{') depth++;
      else if (ch === '}') depth--;
      continue;
    }

    // Outside interpolation: record unescaped backticks (not preceded by backslash)
    if (ch === '`' && content[i - 1] !== '\\') {
      const arr = result.get(line) ?? [];
      arr.push(col);
      result.set(line, arr);
    }
  }

  return result;
};

src/utils/promptSync.ts
mike1858:
```suggestion
  content: string; // The actual prompt content with ${VARIABLE_NAME} placeholders
  /**
   * Line offset of the first content line within the original markdown file.
   * This counts how many lines (including frontmatter and delimiters) appear
   * before the first character of `content`, so we can map content-relative
   * line numbers back to real file line numbers when reporting errors.
   */
  contentLineOffset: number;

mike1858 (approved):
I refactored `findUnescapedBackticks` to be a lot shorter.  I also fixed an issue where the line numbers in the errors were starting from the prompt itself (instead from the beginning of the file).  Otherwise, LGTM.

PR #241, opened 2025-11-30
Fix `$` variable name replacement in system prompts

### LINKED ISSUE #237: Claude Code 2.0.55 unusable with tweakcc 3.1.3
I already tried https://github.com/Piebald-AI/tweakcc#troubleshooting. 

Image

Anytime I type anything when the tweakcc patches are applied, I am interrupted for some reason. I don't know why.

So then I restore the original Claude Code (preserves config.json) option from tweakcc, and everything works normally. 

Image

This was working fine for 2.0.54, which is why I believe it might a regression.

signadou:
Closes #237

### DIFF
--- a/src/utils/patches/systemPrompts.ts
+++ b/src/utils/patches/systemPrompts.ts
@@ -143,7 +143,7 @@ export const applySystemPrompts = async (
       const matchLength = match[0].length;
 
       // Replace the matched content with the interpolated content from the markdown file
-      // Use a replacer function to avoid special replacement pattern interpretation (e.g., $$ -> $)
+      // Use a replacer function to avoid special replacement pattern interpretation (e.g., $$ -> $), see #237
       content = content.replace(pattern, () => interpolatedContent);
 
       // Store the hash of the applied prompt content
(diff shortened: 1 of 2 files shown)


src/utils/patches/systemPrompts.ts
```diff
@@ -143,7 +143,8 @@ export const applySystemPrompts = async (
       const matchLength = match[0].length;
 
       // Replace the matched content with the interpolated content from the markdown file
-      content = content.replace(pattern, interpolatedContent);
+      // Use a replacer function to avoid special replacement pattern interpretation (e.g., $$ -> $)
```
bl-ue:
```suggestion
      // Use a replacer function to avoid special replacement pattern interpretation (e.g., $$ -> $), see #237
```

src/utils/promptSync.ts
```diff
@@ -1272,7 +1272,8 @@ const applyIdentifierMapping = (
 
   for (const [humanName, actualVar] of sortedEntries) {
     const pattern = new RegExp(`\\b${humanName}\\b`, 'g');
-    result = result.replace(pattern, actualVar);
+    // Use a replacer function to avoid special replacement pattern interpretation (e.g., $$ -> $)
```
bl-ue:
```suggestion
    // Use a replacer function to avoid special replacement pattern interpretation (e.g., $$ -> $), see #237
```
Need to fix this duplication some time.

bl-ue (changes requested):

PR #257, opened 2025-12-05
Fix node-lief require in ESM context

mermachine:
Use createRequire to create a require function compatible with ESM. Fixes "Calling require for node-lief in an environment that doesn't expose the require function" error on Node.js 25+ / Linux.

### DIFF
--- a/src/utils/nativeInstallation.ts
+++ b/src/utils/nativeInstallation.ts
@@ -4,8 +4,11 @@
 
 import fs from 'node:fs';
 import { execSync } from 'node:child_process';
+import { createRequire } from 'node:module';
 import type LIEF from 'node-lief';
 import { isDebug } from './misc.js';
+
+const require = createRequire(import.meta.url);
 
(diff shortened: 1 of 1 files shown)


bl-ue:
Cool, thank you!  I wasn't able to reproduce that error on node v25.2.1 with the code from main.  I just used `pnpm build` and then `node dist/index.js`&mdash;are you using different commands?

mermachine:
> Cool, thank you! I wasn't able to reproduce that error on node v25.2.1 with the code from main. I just used `pnpm build` and then `node dist/index.js`—are you using different commands?

i ran npx tweakcc first, which failed, then used npm to build (not pnpm) on WSL/linux !

mermachine:
@bl-ue hey did you want to merge this PR? is ok if not just making sure so I don't close something that's useful ^^

bl-ue:
@mermachine I'm going to close this because @signadou found out a better way to solve it, which involved updating `node-lief`.  Thank you though!

PR #262, opened 2025-12-05
feat: add system prompts for Claude Code v2.0.59

seanGSISG:
## Summary
- Add `prompts-2.0.59.json` to support the latest Claude Code version

## Problem
When running `tweakcc` with Claude Code v2.0.59, system prompts fail to load:
```
✖ Error downloading system prompts:
  Prompts file not found for Claude Code v2.0.59. This version was released within the past day or so and will be supported within a few hours.
```

## Solution
Extract and add the prompts file for v2.0.59.

## Test plan
- [ ] Run `tweakcc` with Claude Code v2.0.59
- [ ] Verify system prompts load successfully (no download error)
- [ ] Verify prompts sync to `~/.tweakcc/system-prompts/`

### DIFF
--- a/src/utils/nativeInstallation.ts
+++ b/src/utils/nativeInstallation.ts
@@ -4,8 +4,12 @@
 
 import fs from 'node:fs';
 import { execSync } from 'node:child_process';
+import { createRequire } from 'node:module';
 import type LIEF from 'node-lief';
 import { isDebug } from './misc.js';
+
+// Create a require function that works in ESM context
+const require = createRequire(import.meta.url);
 
 let liefModule: typeof LIEF | null = null;
 
@@ -15,7 +19,6 @@ let liefModule: typeof LIEF | null = null;
  */
 function getLIEF(): typeof LIEF {
   if (liefModule === null) {
-    // eslint-disable-next-line @typescript-eslint/no-require-imports
     liefModule = require('node-lief');
   }
   return liefModule as typeof LIEF;


bl-ue:
Hi @seanGSISG, thank you for doing this!  I've not had the time lately to get to these.

There's a bit of a procedure here that you need to follow in order to add these files.  First, you need to copy the _old_ prompt JSON file (so in this case for `2.0.58`) and rename it to `2.0.59`.  Then, you run the script on the copy.  If you don't copy the old file first, then all of the names, IDs, and descriptions for the prompt parts in the new file end up as empty strings.    It looks like you might've already figured that part of it.

The second part is diffing the old file (`2.0.58`) and the new file and copying over missing names, IDs, and descriptions from the old file into the new file.  The way the script works, if any prompt changes at all in a new version, the script conservatively avoids copying over any metadata for that prompt because it can't be 100% sure that it's targeting the corresponding prompt in the new file.  (I'll see about adding a similarity check to the script to make this more automated.)  So the name, IDs, descriptions, and all the human-readable variable names are empty for several prompts in [data/prompts/prompts-2.0.59.json](https://github.com/Piebald-AI/tweakcc/pull/262/files#diff-3e996e573947d6866c71eb6b8b90307d8b4d860f8169410cc0ab16ab1bcafb99).

I just merged [a PR](#263) that adds the prompts for 2.0.58, so merge main into your branch, and re-run the script with 2.0.58 as a basis&mdash;there should be fewer diffs as a result.

I appreciate you doing this, so I'm going to improve the script and add docs for how to use it to make it easier.

pwagland:
In the meantime, #273 has been merged which I believe supersedes this?

bl-ue:
@seanGSISG I'm going to close this because @mike1858 has taken care of the prompts for 59 and the 10 versions that have been released since.  Thank you though!

PR #282, opened 2025-12-12
fix: prefer latest bunx cached claude-code version when patching

y0usaf:
## Summary

When multiple bunx cached versions of Claude Code exist, tweakcc now selects and patches the latest version instead of whichever appears first in the filesystem.

## Problem

Users with multiple bunx cached Claude Code versions would have older versions patched instead of the newest, preventing customizations from being applied to their current installation.

## Solution

Added semantic version parsing and sorting to prioritize bunx cache entries by version in descending order:

- **extractVersionFromPath()**: Parses semantic versions from bunx cache paths (e.g., extracts [2, 0, 67] from ~/.local/share/bun/install/cache/@anthropic-ai/claude-code@2.0.67@@@hash)
- **compareVersions()**: Compares two semantic versions using standard major.minor.patch comparison
- **Sorted search**: Before iterating through installation paths, sorts them so bunx cache entries are checked newest-first

## Testing

Verified with real bunx cache containing versions 2.0.54 through 2.0.67 - tweakcc now correctly selects and patches 2.0.67.

### DIFF
--- a/src/tests/bunxVersionSorting.test.ts
+++ b/src/tests/bunxVersionSorting.test.ts
@@ -17,7 +17,9 @@ describe('extractVersionFromPath', () => {
 
   it('should return null for non-bunx paths', () => {
     expect(
-      extractVersionFromPath('/usr/local/lib/node_modules/@anthropic-ai/claude-code')
+      extractVersionFromPath(
+        '/usr/local/lib/node_modules/@anthropic-ai/claude-code'
+      )
     ).toBeNull();
   });
 


src/installationDetection.ts
```diff
@@ -238,6 +238,32 @@ async function findClijsFromExecutablePath(
   return null;
 }
 
+/**
+ * Extracts version from a bunx cache path string.
+ * Bunx cache paths follow the pattern: .../@anthropic-ai/claude-code@VERSION@@@HASH
+ * Returns [major, minor, patch] as numbers or null if pattern not found.
+ */
+function extractVersionFromPath(pathStr: string): [number, number, number] | null {
+  const match = pathStr.match(/@anthropic-ai[\\\/]claude-code@(\d+)\.(\d+)\.(\d+)/);
```
bl-ue:
```suggestion
  const match = pathStr.match(/@anthropic-ai[\\/]claude-code@(\d+)\.(\d+)\.(\d+)/);
```

PR #353, opened 2026-01-09
fix(thinkingVisibility): support v2.1.x native binary format

gulp:
## Summary

Fixes `thinkingVisibility` patch for Claude Code v2.1.x native binaries.

The current patch fails with:
```
patch: thinkingVisibility: failed to find thinking visibility pattern
```

### Root cause

Two issues:
1. The v2.1.x minified format has an opening brace after `case"thinking":` that the current regex doesn't expect
2. Full visibility requires patching TWO locations, not just one

### Solution

**Patch 1: Case statement** - prevents thinking blocks from being skipped
```javascript
// Original
case"thinking":{if(!J&&!I)return null
// Patched  
case"thinking":{if(!1&&!1)return null
```

**Patch 2: FbH collapsed view** - shows expanded view instead of "(ctrl+o to expand)"
```javascript
// Original
if(!(A||L))return p0H
// Patched
if(!(1||1))return p0H
```

### Key changes

- Add support for new v2.1.x format with opening brace after case
- Add second patch for FbH collapsed view function  
- Use flexible regex to handle variable name changes between versions
- Maintain backwards compatibility with older format
- Patch FIRST occurrence only (binary has duplicate code sections)

## Test plan

- [x] Tested on Claude Code v2.1.2 native binary (Linux x64)
- [x] All existing tests pass
- [x] Build succeeds

## Note

Other patches also need updating for v2.1.x (thinker verbs, context limit, etc.) - those are separate issues.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the "thinking visibility" behavior to more reliably reveal thinking content across multiple version formats and patterns.
  * Applies a coordinated two-step update process to expand collapsed views more safely, using incremental changes to preserve correctness.
  * Better error/warning messages when only partial updates can be applied, increasing diagnostic clarity for edge cases.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/.claude/current-session
+++ b/.claude/current-session
@@ -1,0 +1,1 @@
+ad0db82b-f95e-4189-b804-edaf9cd6bd80
--- a/src/patches/thinkingVisibility.ts
+++ b/src/patches/thinkingVisibility.ts
@@ -120,12 +120,12 @@ export const writeThinkingVisibility = (oldFile: string): string | null => {
 
   // Patch 1: Case statement visibility
   const caseLocation = getCaseStatementLocation(oldFile);
-  if (caseLocation) {
+  if (caseLocation?.identifiers) {
     let replacement: string;
 
-    if (caseLocation.identifiers!.length === 4) {
+    if (caseLocation.identifiers.length === 4) {
       // Old format: remove if/return and set isTranscriptMode to true
-      replacement = `${caseLocation.identifiers![0]}${caseLocation.identifiers![1]}${caseLocation.identifiers![2]}${caseLocation.identifiers![3]}`;
+      replacement = `${caseLocation.identifiers[0]}${caseLocation.identifiers[1]}${caseLocation.identifiers[2]}${caseLocation.identifiers[3]}`;
     } else {
       // New format: replace variable names with 1
       replacement = `case"thinking":{if(!1&&!1)return null`;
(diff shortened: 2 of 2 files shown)


gulp:
Closing in favor of #352 which provides a more comprehensive fix for all v2.1.x patches. Tested and confirmed working. Thanks @basekevin!

bl-ue:
Hi @gulp!  Thank you very much for fixing this!  You're right, we already have a PR in to fix all the patching errors&mdash;I'm sorry for the duplicated effort!

Thank you for testing #352 as well!  I'll merge it and make a new release.

gulp:
Did you guys figure out how to display thinking blocks as dim and _italic_ like once before the update? Or am I the only one bothered by it?

image

bl-ue:
@gulp You're right, that's terrible!  I just added a patch that fixed that for me.  Will you please try checking out #369's `thinking-style-patch` branch and see if it fixes it for you?

bl-ue:
@gulp By the way they seem to have fixed this bug in CC.  It's a bit different than what tweakcc does but it's probably fine for you.  The first screenshot shows what tweakcc does (what CC used to do before it broke), and the second shows what CC now does by default.

| tweakcc |  CC 2.1.26+ builtin |
| ------ | ------ |
| image | image |

PR #372, opened 2026-01-13
fix: make userMessageDisplay opt-in to prevent CC 2.1.6+ input bug

treygoff24:
1 ## Summary
   2 
   3 The `userMessageDisplay` patch causes an "insta-ESC" bug on Claude Code 2.1.6 where user input is immediately cancelled/escaped. This makes Claude Code completely unusable when the patch is applied.
   4 
   5 ### Root Cause
   6 
   7 The patch modifies the component structure for rendering user messages (wrapping in a Box component, changing the React element tree). This structural change appears to interfere with input handling in CC 2.1.6+.
   8 
   9 ### The Fix
  10 
  11 Make `userMessageDisplay` **opt-in** instead of always-on:
  12 
  13 - Add `enabled?: boolean` field to `UserMessageDisplayConfig` type
  14 - Default to `enabled: false` in `DEFAULT_SETTINGS`
  15 - Only apply the patch when `enabled === true` (strict equality check)
  16 - Show a helpful migration warning for users who have custom styling configured but haven't explicitly enabled the feature
  17 
  18 This allows users to continue using all other tweakcc features (themes, thinking verbs, etc.) safely, while making `userMessageDisplay` opt-in for those willing to test it on their CC version.
  19 
  20 ### Why `enabled === true` (not `!== false`)
  21 
  22 Existing user configs won't have the `enabled` key, so:
  23 - `undefined !== false` → `true` ❌ would still apply broken patch
  24 - `undefined === true` → `false` ✅ safely skips the patch
  25 
  26 ## Test Plan
  27 
  28 - [x] TypeScript compiles without errors
  29 - [x] All existing tests pass (132 passed)
  30 - [x] Tested on Claude Code 2.1.6 - confirmed working
  31 - [x] Other tweakcc features (themes, thinking verbs) still work correctly
  32 - [x] Migration warning displays for users with custom styling
  33 
  34 ## Files Changed
  35 
  36 - `src/types.ts` - Added `enabled?: boolean` to `UserMessageDisplayConfig`
  37 - `src/defaultSettings.ts` - Set `enabled: false` as default
  38 - `src/patches/index.ts` - Check for explicit opt-in + migration warning
  39 
  40 ---
  41 
  42 🤖 Generated with [Claude Code](https://claude.ai/code)

## Summary by CodeRabbit

* **Bug Fixes**
  * User message display customization is now disabled by default to improve compatibility. Enable it explicitly in settings if custom styling is needed.

* **Improvements**
  * Added warning notifications alerting users when message customization is configured but disabled.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/defaultSettings.ts
+++ b/src/defaultSettings.ts
@@ -731,6 +731,7 @@ export const DEFAULT_SETTINGS: Settings = {
     reverseMirror: true,
   },
   userMessageDisplay: {
+    enabled: false,
     format: ' > {} ',
     styling: [],
     foregroundColor: 'default',
--- a/src/patches/index.ts
+++ b/src/patches/index.ts
@@ -540,27 +540,42 @@ export const applyCustomization = async (
   if ((result = writeThinkerSymbolMirrorOption(content, config.settings.thinkingStyle.reverseMirror)))
     content = result;
 
-  // Apply user message display customization
-  if (config.settings.userMessageDisplay) {
+  // Apply user message display customization (opt-in only due to CC 2.1.6+ compatibility issues)
+  const displayConfig = config.settings.userMessageDisplay;
+  if (displayConfig?.enabled === true) {
     if (
       (result = writeUserMessageDisplay(
         content,
-        config.settings.userMessageDisplay.format,
-        config.settings.userMessageDisplay.foregroundColor,
-        config.settings.userMessageDisplay.backgroundColor,
-        config.settings.userMessageDisplay.styling.includes('bold'),
-        config.settings.userMessageDisplay.styling.includes('italic'),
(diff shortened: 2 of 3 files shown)


bl-ue:
Hi @treygoff24!  The "inst-esc" issue you described sounds exactly like what happens when tweakcc accidentally breaks some Claude Code internals.  It's because there's actually an error occurring when you send a message, but Claude Code doesn't normally display.  To see the actual error, will you please
1. run `claude --debug` (with your tweakcc patches applied),
2. open the log file that will be indicated in the output,
3. send a message to trigger the error,
4. look for a JavaScript error towards the end of the log file, and
5. send that log file here?

PR #408, opened 2026-01-24
Gate automatic updates on published prompt snapshots

### LINKED ISSUE #170: Feature: prevent auto-update to unsupported versions
Hey, I was wondering if it would be possible to maintain auto-update for Claude Code enabled but not allow it to go to a version beyond tweakcc's current support.
This will allow the users to maintain latest features automatically but also never losing tweakcc (for cases where you are reliant on the system prompts modifications).

Thanks

mike1858:
## Why

Automatic Claude Code updates can move a customized installation to a version whose prompt snapshot TweakCC has not published yet. This adds an opt-in gate at the automatic updater's final version decision, while leaving manual updates available.

Closes #170.

Depends on #987 for module-aware native read/write support. This PR targets its branch so the updater changes can be reviewed separately.

## What changed

- Add the Misc toggle `settings.misc.preventUpdateToUnsupportedVersions`, defaulting to `false`, and include it in the pre-apply plan only when enabled and selected.
- Require HTTP 200 from the target version's prompt snapshot in `Piebald-AI/tweakcc`. Missing snapshots, transport failures, and checks exceeding five seconds block the automatic update. Validate the version before constructing the URL.
- For the split native updater, pass a guard from the automatic caller through to the resolved installer target, after canary selection and version caps. Keep guarded work separate from the manual installer's single-flight result. Locate all required edits before staging either module.
- For the legacy npm updater, pin local, global, and fallback installation commands to the exact checked version instead of resolving a moving `stable`/`latest` tag again.
- Apply native chunk edits through the module-aware writer, parse ESM chunks before writing, and report chunk-only changes as applied. Disabling or filtering out the guard restores the original without routing an unchanged modern graph through the legacy repacker. If corpus extraction fails but ordinary native extraction succeeds, report the guard as failed and continue unrelated customizations without using the npm-only matcher.

## Compatibility and limits

Verified source layouts are Claude Code **2.1.20 npm** and **2.1.261 native**. Unknown layouts fail patch matching rather than receiving a partial guard. Native executable testing was performed on Linux x64.

A published prompt snapshot is not a compatibility guarantee for every patch in the locally installed TweakCC release. Manual `claude update` / `claude install` calls and external package-manager updates remain unchanged. A successful update replaces the patched executable, so reapply TweakCC afterward to restore this guard and other customizations.

## Validation

- `pnpm lint` — passing.
- `pnpm test` — 576 passed, 5 skipped.
- `pnpm build` — passing.
- `pnpm exec prettier --check src` — passing.
- `pnpm exec vitest run src/patches/preventUnsupportedUpdates.test.ts` — 20 tests covering final target selection, HTTP/network/deadline failure, manual calls, single-flight isolation, missing/ambiguous layouts, identifier collisions, and exact npm version pinning.
- `pnpm exec vitest run src/patches/preventUnsupportedUpdatesFallback.test.ts src/patches/preventUnsupportedUpdates.test.ts` — 24 passed, including recoverable native extraction, unrelated patch application, no native-to-npm matcher fallback, and failure of both extractors.

Ran the built CLI with `TWEAKCC_CONFIG_DIR` and `TWEAKCC_CC_INSTALLATION_PATH` pointing exclusively to disposable fixture paths:

```sh
node dist/index.mjs --apply --yes --patches prevent-unsupported-updates
```

```text
  Misc Configurable:
    ✓ Prevent unsupported updates — Native/npm auto-updates require a published prompt snapshot

Customizations applied successfully!
```

The patched 2.1.261 executable runs `--version`. Exactly **2 of 1,818** module sources change. Disabling or filtering out the guard restores a byte-identical original; reapplying produces byte-identical patched output. The patched 2.1.20 npm source passes module-mode `node --check` and prints `2.1.20 (Claude Code)` with `--version`.

### Shipped updater behavior

Executed the extracted, shipped updater function bodies before and after patching, with network, filesystem, and installer effects stubbed. These are behavioral checks, **not real installations**:

```text
Before / missing snapshot: installerCalls=1, skipped=false, target=2.1.262, supportChecks=0
After / missing snapshot: installerCalls=0, skipped=true, target=2.1.262, supportChecks=1
After / published snapshot: installerCalls=1, skipped=false, target=2.1.262, supportChecks=1
After / GitHub failure: installerCalls=0, skipped=true, target=2.1.262, supportChecks=1
After / manual install: installerCalls=1, skipped=false, target=2.1.262, supportChecks=0
After / policy-capped target: installerCalls=0, skipped=true, target=2.1.263, supportChecks=1
After / canary target: installerCalls=0, skipped=true, target=2.1.265, supportChecks=1
```

Legacy npm command construction uses the checked version:

```text
Before / npm-global: npm install -g @anthropic-ai/claude-code; supportChecks=0
Before / npm-local: npm install @anthropic-ai/claude-code@stable; supportChecks=0
After / npm-global: npm install -g @anthropic-ai/claude-code@2.1.21; supportChecks=1
After / npm-local: npm install @anthropic-ai/claude-code@2.1.21; supportChecks=1
```

Both fallback branches also use `@2.1.21`; a missing snapshot produces no installer call in any of the four branches. Live HEAD requests returned HTTP 200 for the published 2.1.261 snapshot and HTTP 404 for 99.0.0. No installed Claude Code executable was modified.

## Summary by CodeRabbit

- **New Features**
  - Added an optional setting to prevent automatic updates to unsupported versions.
  - Automatic native and npm updates now verify prompt snapshot availability before proceeding; failed, unavailable, or timed-out checks block the update.
  - Manual updates and package-manager updates remain unchanged.
  - Other customizations can still be applied if the update protection check cannot be installed.

- **Documentation**
  - Added guidance for configuring and using update protection.

- **Tests**
  - Added coverage for configuration, update safeguards, module parsing, and installation layouts.


src/patches/preventUnsupportedUpdates.ts
```diff
@@ -0,0 +1,139 @@
...
+  const nextFunc = location.identifiers![5];
+
+  // Construct the replacement with the tweakcc version check injected
+  // The check wraps the version fetch to check if tweakcc supports the version.
+  // If the prompts file doesn't exist (404), it returns the current version to block the update.
+  //
+  // TEST MODE: Spoofing version to 99.0.0 to test blocking behavior
+  // TODO: Remove test code before merging!
+  const tweakccVersionCheck = `${versionVar}=await(async()=>{let v=await ${fetchFunc}(${channelVar});require("fs").appendFileSync("/tmp/tweakcc-test.log","[tweakcc] Fetched version: "+v+"\\n");v="99.0.0";require("fs").appendFileSync("/tmp/tweakcc-test.log","[tweakcc] Spoofed to: "+v+"\\n");try{const r=await fetch(\`https://raw.githubusercontent.com/Piebald-AI/tweakcc/refs/heads/main/data/prompts/prompts-\${v}.json\`,{method:'HEAD'});require("fs").appendFileSync("/tmp/tweakcc-test.log","[tweakcc] GitHub status: "+r.status+"\\n");if(!r.ok){require("fs").appendFileSync("/tmp/tweakcc-test.log","[tweakcc] BLOCKING update!\\n");return ${currentVersionVar};}}catch(e){require("fs").appendFileSync("/tmp/tweakcc-test.log","[tweakcc] Error: "+e+"\\n")}return v;})(),`;
```
bl-ue: Forgot to remove?

bl-ue (dismissed):

src/patches/preventUnsupportedUpdates.ts
mike1858: Oops, I guess I didn't read the comment.  :laughing:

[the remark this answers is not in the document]
mike1858: @coderabbitai Oops, I guess I didn't read the comment. 😆

mike1858: Done.

georpar:
@mike1858 Not working for me:

```
patch: preventUnsupportedUpdates: failed to find auto-updater pattern
```

Once you've fixed that (and added the 2.1.20 prompts to the tweakcc repo), to test if the patch is working properly, you can modify the `preventUnsupportedUpdates` patch locally to point it to https://github.com/georpar/tweakcc (doesn't have 2.1.20 prompts), downgrade CC to 2.1.19.  When you're done testing, point it back to Piebald-AI/tweakcc and test again, to make sure updating works when the latest system prompts are supported by tweakcc.

mike1858:
@georpar Can you please try it now and mark it as ready for review if it works?  It _should_ work.

georpar:
The code looks good, and it works with npm installations, but it doesn't work with native installations.  Needs to be investigated by @mike1858 or me.

mike1858:
@coderabbitai review

src/patches/index.ts
```diff
@@ -616,8 +647,19 @@ export const applyCustomization = async (
...
 
-    const claudeJsBuffer =
-      await extractClaudeJsFromNativeInstallation(pathToExtractFrom);
+    // Only this explicitly selected feature opts into module-aware application;
+    // it must patch the actual installer chunk rather than a tiny import stub.
+    if (needsNativeGuard) {
+      nativeCorpus =
+        await extractClaudeJsModulesFromNativeInstallation(pathToExtractFrom);
+      if (!nativeCorpus)
+        throw new Error('Failed to extract native module corpus');
```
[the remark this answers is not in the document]
mike1858: Fixed in e02452c. Recoverable corpus-extraction failures now fall back to ordinary native extraction, report the update guard as failed, and leave unrelated patches available. Native fallback explicitly cannot invoke the npm-only matcher. Four pipeline regression tests cover unrelated patch application, guard-only failure without repacking, failure of both extractors, and the disabled setting; the full suite passes (576 tests, 5 skipped).

mike1858:
@coderabbitai review

mike1858:
The encoding finding uses the older enum, but the proposed change is incorrect for Bun 1.4.1, which is the runtime in the verified Claude Code 2.1.261 native artifact.

At pinned Bun revision `4661e494f052c83c80dade1318e5710238340be6`:

- [The enum](https://github.com/oven-sh/bun/blob/4661e494f052c83c80dade1318e5710238340be6/src/standalone_graph/StandaloneModuleGraph.rs#L407-L419) explicitly defines `Binary = 0`, `Latin1 = 1`, and `Utf16 = 2`. Its comment explains that UTF-16 reuses the value of the **never-written** former `Utf8` variant.
- [The runtime decoder](https://github.com/oven-sh/bun/blob/4661e494f052c83c80dade1318e5710238340be6/src/standalone_graph/StandaloneModuleGraph.rs#L696-L727) sends `Binary` through `clone_utf8`, `Latin1` through the Latin-1 path, and `Utf16` through `utf16_units` / `create_static_external_utf16`.

`nativeSource` therefore intentionally decodes tag 2 as UTF-16LE. Changing it to UTF-8 would corrupt those source modules. The module replacement writer uses tag 0 for its UTF-8 output to avoid this version-dependent tag ambiguity.

Commit afb8f86 preserves the correct behavior, adds the pinned citation beside the decoder, and adds three pipeline regression cases for tags 0/1/2, including accented characters and a supplementary-plane character. All 583 tests pass (5 skipped), along with lint, build, and formatting.

@coderabbitai Please reassess the encoding finding against these pinned runtime sources and review the new regression coverage. Do not apply the suggested UTF-8 decoding change for tag 2.

mike1858:
@coderabbitai review

PR #416, opened 2026-01-27
Prompts for 2.1.20

mike1858:
@bl-ue **Note:** Anthropic refactored the prompts a lot and cut them up in to multiple pieces.  I reviewed it carefully, but when you review it, you should review it more carefully than usual.


mike1858:
@coderabbitai review.

PR #518, opened 2026-02-14
Fix patchesAppliedIndication for CC 2.1.42+ ternary minHeight

zigazaga4:
## Summary
- Claude Code 2.1.42 changed `minHeight:9` to a ternary `minHeight:X?13:9` in the header component
- The hardcoded regex `minHeight:9` in PATCH 4 and PATCH 5 no longer matches, causing `patchesAppliedIndication` to fail
- Updated both regexes to handle ternary expressions (`VAR?NUM:NUM`) as well as plain numbers
- Preserved the original minHeight value dynamically in the replacement instead of hardcoding `9`

## Test plan
- [x] `pnpm build` passes
- [x] `pnpm lint` passes  
- [x] `pnpm test` — all 162 tests pass
- [x] Tested `--apply` on Claude Code 2.1.42 — `patchesAppliedIndication` now shows ✓

## Release Notes

* **Bug Fixes**
  * Fixed layout rendering to properly support dynamic height values in addition to fixed values, ensuring correct display when height is computed rather than static.

### DIFF
--- a/src/patches/patchesAppliedIndication.ts
+++ b/src/patches/patchesAppliedIndication.ts
@@ -67,8 +67,9 @@ const applyIndicatorViewPatch = (
   textComponent: string,
   chalkVar: string
 ): { content: string; closingParenIndex: number } | null => {
-  // 1. Find alignItems:"center",minHeight:9,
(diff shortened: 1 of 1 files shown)


bl-ue:
Thank you for this fix!

PR #564, opened 2026-02-28
Fix agentsMd patch syntax error on CC v2.1.62

viceversaco:
## Summary

- v2.1.62 restructured the CLAUDE.md reading function — it now calls `readFileSync` directly inside `try{}` and catches `ENOENT`, instead of guarding with `existsSync`/`statSync` before reading
- The primary pattern (Step 3) no longer matches, causing the fallback pattern (`}}{catch}`) to inject an `else if` after a `return` statement, producing invalid JavaScript (`return {...} else if (...)`)
- This crashes the Claude binary on startup with `TypeError: Expected CommonJS module to have a function wrapper`

## Fix

Adds a new catch-based pattern (Step 4) between the existing primary and fallback patterns that matches the ENOENT handler:

```
if(errVar==="ENOENT"||errVar==="EISDIR")return null;
```

and wraps it with the alt-name check, matching the same approach used by the primary pattern.

## Backward compatibility

| CC Version | Primary (Step 3) | Catch (Step 4, new) | Fallback (Step 5) |
|------------|-------------------|---------------------|-------------------|
| v2.1.29    | **matches**       | no match            | skipped           |
| v2.0.0     | no match          | no match            | **matches**       |
| v2.1.62    | no match          | **matches**         | skipped           |

## Test plan

- [x] Verified new pattern matches v2.1.62 source correctly
- [x] Verified primary pattern still matches v2.1.29 structure (not affected)
- [x] Verified fallback pattern still matches v2.0.0 structure (not affected)
- [x] `node --check` passes on patched v2.1.62 JS (was failing before with `SyntaxError: Unexpected token 'else'`)
- [x] All 162 existing tests pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced compatibility with Code Clarity versions 2.1.29 through 2.1.62+ with improved fallback logic
  * Improved error handling and recovery when primary agent configuration files are unavailable or inaccessible
  * Added support for automatic fallback to alternative agent configuration sources when primary paths are not accessible
  * Maintained backward compatibility while expanding coverage across different version variants

### DIFF
--- a/src/patches/agentsMd.ts
+++ b/src/patches/agentsMd.ts
@@ -94,6 +94,40 @@ import { showDiff } from './index';
  *    return null;
  *  }
  * ```
+ *
+ * CC 2.1.62 (reads file first, catches ENOENT):
+ * ```
+ *  function kPA(H, $) {
+ *    try {
+ *      let L = mH().readFileSync(H, { encoding: "utf-8" }),
+ *        I = fM.extname(H).toLowerCase();
+ *      if (I && !sC1.has(I))
+ *        return T(`Skipping non-text file in @include: ${H}`), null;
+ *      let { content: D, paths: B } = eC1(L), f = D;
+ *      if ($ === "AutoMem" || $ === "TeamMem") { ... }
+ *      return { path: H, type: $, content: f, globs: B };
+ *    } catch (A) {
+ *      let L = A.code;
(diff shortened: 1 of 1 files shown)


viceversaco:
totally didn't mean to make this pr. vibes man lol

PR #572, opened 2026-03-01
fix: make model customizations configurable and update models to latest

liafonx:
## Why

Model customizations had no config/TUI opt-out.

## What changed

- Added `settings.misc.enableModelCustomizations` (default `true`) and wired Misc UI toggle.
- Moved `model-customizations` and `show-more-items-in-select-menus` from Always Applied to Misc Configurable, gated by the new setting.
- Updated model catalog (added Sonnet 4.6, Haiku 4.5) and updated opusplan descriptions from 4.5 to 4.6.
- Widened modelSelector lookback window (600→1500) and generalized declaration pattern (`let|var|const`) for minifier variance.
- Preserved wrapper function semantics in opusplan1m selector injection.

## Validation

- `npx vitest run src/` — 168 passed, 4 skipped
- `pnpm lint` — clean
- Clean apply on fresh `brew reinstall claude-code`

## Reviewer checklist

- TUI toggle and config backfill for `enableModelCustomizations` work end-to-end
- Patch gating behavior matches setting state (`true`/`false`/unset)
- No regressions in existing patch ordering/grouping

## Summary by CodeRabbit

* **New Features**
  - Added a new setting to toggle model customizations on/off (enabled by default), controlling whether all available models are displayed
  - Updated model list with newer Claude versions: Opus 4.6, Sonnet 4.6, and Haiku 4.5
  - Added support for opusplan[1m] model variant

* **Tests**
  - Added test coverage for model customization toggle functionality

### DIFF
--- a/data/prompts/prompts-2.1.64.json
+++ b/data/prompts/prompts-2.1.64.json
@@ -1,0 +1,3493 @@
+{
+  "version": "2.1.64",
+  "prompts": [
+    {
+      "name": "Agent Prompt: Agent creation architect",
+      "id": "agent-prompt-agent-creation-architect",
+      "description": "System prompt for creating custom AI agents with detailed specifications",
+      "pieces": [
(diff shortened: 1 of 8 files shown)


bl-ue:
Hi @liafonx, thank you for these fixes!

The `UNKNOWN_*` almost always means a bug in our extraction of the system prompts, not a bug in tweakcc.  What version of Claude Code are you using?  I found that the prompts for 2.1.16 and 2.1.17 both have one empty identifier which was getting mapped to `UNKNOWN_*`, but those are fairly old versions, and that still doesn't explain how the counter got to `3` in `UNKNOWN_3` since there was only 1 missing identifier in each prompt, not 3.  (#575)

Everything else&mdash;the model selector changes, the configuration option for them, the opusplan[1m] changes&mdash;looks great; thank you!

liafonx:
> Hi @liafonx, thank you for these fixes!
> 
> The `UNKNOWN_*` almost always means a bug in our extraction of the system prompts, not a bug in tweakcc. What version of Claude Code are you using? I found that the prompts for 2.1.16 and 2.1.17 both have one empty identifier which was getting mapped to `UNKNOWN_*`, but those are fairly old versions, and that still doesn't explain how the counter got to `3` in `UNKNOWN_3` since there was only 1 missing identifier in each prompt, not 3. (#575)
> 
> Everything else—the model selector changes, the configuration option for them, the opusplan[1m] changes—looks great; thank you!

Hi @bl-ue , I am using v2.1.63. The reason for that fix is that when I enable opusplan mode (even without [1m]), then switch to plan mode and send a message, my request gets silently dropped with an error:

> claude -d  --model opusplan --permission-mode plan --output-format text -p --verbose  hello
{"type":"result","subtype":"error_during_execution","duration_ms":0,"duration_api_ms":0,"is_error":true,"num_turns":0,"stop_reason":null,"session_id":"94495963-b187-4879-a0c7-251da0247a92","total_cost_usd":0,"usage":{"input_tokens":0,"cache_creation_input_tokens":0,"cache_read_input_tokens":0,"output_tokens":0,"server_tool_use":{"web_search_requests":0,"web_fetch_requests":0},"service_tier":"standard","cache_creation":{"ephemeral_1h_input_tokens":0,"ephemeral_5m_input_tokens":0},"inference_geo":"","iterations":[],"speed":"standard"},"modelUsage":{},"permission_denials":[],"uuid":"80c597fb-6234-454b-9a64-f786071d47e0","errors":["UNKNOWN_3 is not defined"]}

Strangely, this only happens when I switch to plan mode. Even after switching back, changing the model from opus to sonnet, the issue persists. Only restarting Claude Code resolves it. After investigating with Claude by checking logs and tweaked binary, I found that was the cause of the issue. However, as you said, since I haven’t examined the entire injection system, The `UNKNOWN_*` might be a smokescreen. Still, after applying that patch, the issue was resolved.

bl-ue:
@liafonx Okay, thank you, that's helpful.  To help me debug, will you please trying using vanilla tweakcc without your patch and produce the broken binary, and then run `npx tweakcc@latest unpack broken.js` and send the broken JS file here?

Will you please also revert the `UNKNOWN_*` changes so that I can merge this PR?  I think your change might be a good one, but I'd first like to understand the root cause here, and then maybe implement a warning or something when it happens in the future.

liafonx:
Hi [@bl-ue](https://github.com/bl-ue), done — reverted the UNKNOWN_* fallback in systemPromptSync.ts as requested. The PR now only contains the model customizations toggle and model catalog updates.

Here is the reproduction as requested:

**Environment**: CC v2.1.63, vanilla `npx tweakcc@latest` (4.0.10), macOS arm64

**Steps**:
```
brew reinstall claude-code
npx --yes tweakcc@latest --apply
claude -d --model opusplan --permission-mode plan --output-format text -p --verbose hello
```

**Error output**:
```json
{"type":"result","subtype":"error_during_execution","is_error":true,"errors":["UNKNOWN_3 is not defined"]}
```

**Broken JS** (extracted via `npx tweakcc@latest unpack broken.js`):
https://gist.github.com/liafonx/153f74c11df924c11e45d8dfe9460fa0

The gist contains the full 11MB patched JS. `grep UNKNOWN` shows both `UNKNOWN_3` and `UNKNOWN_4` present (4 occurrences total).

PR #579, opened 2026-03-04
Fix AGENTS.md patch replacing CLAUDE.md content instead of falling back

TyceHerrman:
## Summary
- The `agentsMd` patch injected fallback logic at the wrong location — before `return { content }`, which runs *after* CLAUDE.md is successfully read, unconditionally replacing its content with AGENTS.md
- Moved injection to the early `return null` (after `existsSync`/`statSync` check) so alternatives are only tried when CLAUDE.md doesn't exist
- Added 7 unit tests covering fallback behavior, CLAUDE.md preservation, recursion guard, and error cases

## Test plan
- [x] `pnpm build` passes (no type errors)
- [x] `pnpm test` passes (169 tests, including 7 new ones)
- [ ] Manual: apply tweakcc to a Claude Code installation, verify that in a project with both `CLAUDE.md` and `AGENTS.md`, the `CLAUDE.md` content is used (not replaced)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of missing configuration files with automatic fallback to alternative file names.

* **Tests**
  * Added comprehensive test coverage for file routing and fallback logic.

### DIFF
--- a/src/patches/agentsMd.test.ts
+++ b/src/patches/agentsMd.test.ts
@@ -1,0 +1,65 @@
+import { describe, it, expect } from 'vitest';
+import { writeAgentsMd } from './agentsMd';
+
+const mockFunction =
+  'function _t7(A,q){try{let K=x1();' +
+  'if(!K.existsSync(A)||!K.statSync(A).isFile())return null;' +
+  'let Y=UL9(A).toLowerCase();' +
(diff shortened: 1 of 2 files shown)


bl-ue (approved):
Thank you!

PR #585, opened 2026-03-05
Fix: Support Claude Code 2.1.69 module path

### LINKED ISSUE #583: New `Error: Could not extract JS from native binary` for 2.1.69
Tried a fresh install of both cc and tweakcc to no avail. using tweakcc 4.0.10 on macos.

### LINKED ISSUE #584: Failed to extract claude.js from native installation on Claude Code 2.1.69 — module renamed
## Problem

`npx tweakcc --apply` fails on Claude Code **2.1.69** with:

```
Error: Failed to extract claude.js from native installation
```

## Root Cause

Claude Code 2.1.69 renamed the main embedded Bun module from `/$bunfs/root/claude` to `/$bunfs/root/src/entrypoints/cli.js`.

The module name matcher in `nativeInstallation` only checks:

```js
function l(e) {
  return e.endsWith(`/claude`) || e === `claude` || e.endsWith(`/claude.exe`) || e === `claude.exe`
}
```

This no longer matches the new name.

| CC Version | Module 0 | Match? |
|---|---|---|
| 2.1.68 | `/$bunfs/root/claude` | ✅ Yes |
| 2.1.69 | `/$bunfs/root/src/entrypoints/cli.js` | ❌ No |

## Suggested Fix

Update the matcher to also accept `cli.js` / `entrypoints/cli.js`:

```js
function l(e) {
  return e.endsWith(`/claude`) || e === `claude`
    || e.endsWith(`/claude.exe`) || e === `claude.exe`
    || e.endsWith(`/src/entrypoints/cli.js`)
}
```

## Environment

- tweakcc: 4.0.10
- Claude Code: 2.1.69
- Platform: macOS arm64

zigazaga4:
tweakcc failed to extract claude.js from Claude Code 2.1.69 because the module naming convention changed from '`//root/claude`' to '`//root/src/entrypoints/cli.js`'.

This fix extends `isClaudeModule()` in src/nativeInstallation.ts to recognize both the legacy and new path patterns, enabling tweakcc to work with Claude Code 2.1.69+.

**Changes:**
- Added conditions to detect modules ending with '/src/entrypoints/cli.js'
- Added condition to detect modules named exactly 'src/entrypoints/cli.js'

Fixes #584
Fixes #583

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of Claude CLI module variants so the CLI entrypoint is reliably recognized across more installation layouts.

* **Documentation**
  * Added changelog entry noting support for the Claude Code 2.1.69 module path.


bl-ue:
Haha, I'm sorry about this mess.  GitHub wasn't letting merge the PR and I couldn't figure out why at first, but then I found https://www.githubstatus.com/incidents/g5gnt5l5hf56, so I just temporarily disabled the CI checks requirement.  Thank you @zigazaga4!

zigazaga4:
**Update: This fix is incomplete and causes issues.**

The `isClaudeModule()` change correctly allows tweakcc to **extract** the JS from the 2.1.69 binary, but applying patches afterwards breaks Claude Code — it stops responding to messages entirely (blank output, no response).

The root issue is that Claude Code 2.1.69 restructured its internals (module path changed from `/$bunfs/root/claude` to `/$bunfs/root/src/entrypoints/cli.js`), and the existing pattern-matching patches (themes, verbose property, thinking verbs, etc.) are hitting wrong targets in the new code structure, silently corrupting the output.

**This PR should NOT be merged as-is.** The extraction fix alone is not enough — all pattern-matching patches need to be audited and updated for 2.1.69's new code layout.

bl-ue:
@zigazaga4 I just figured that out too... I tested that patched CC launches on Windows, but I didn't send a message until after I merged.  I'll investigate.

PR #592, opened 2026-03-06
fix: apply defaultToolset fallback when persisted state has no toolset field

y0usaf:
## Summary

- When Claude Code loads persisted app state from a previous session, the state may not include a `toolset` field (saved before the toolset patch was first applied)
- This caused `currentToolset` to be `undefined`, falling through to the unfiltered `else`-branch in `writeToolFetchingUseMemo`, exposing all tools on startup
- Users had to Shift+Tab through permission modes to trigger `writeModeChangeUpdateToolset`, which explicitly sets the toolset in state — only then did the restriction activate

## Fix

Pass `defaultToolset` into `writeToolFetchingUseMemo` and emit `?? defaultToolset` in the generated code:

```js
// Before
let currentToolset = appStateSelector(state => state.toolset);

// After  
let currentToolset = appStateSelector(state => state.toolset) ?? "default";
```

This ensures the configured default toolset is active from the very first render, regardless of what persisted state contains.

## Summary by CodeRabbit

* **New Features**
  * App now accepts and honors a provided default toolset on first render, allowing a preselected tool configuration.

* **Bug Fixes**
  * Improved handling when persisted toolset data is missing: falls back to the provided default (or undefined) to avoid uninitialized state.
  * Mode-specific toolset behavior now applies conditionally when both a plan-mode toolset and a default toolset are present.

### DIFF
--- a/src/patches/toolsets.ts
+++ b/src/patches/toolsets.ts
@@ -331,7 +331,8 @@ if (toolsets.hasOwnProperty(currentToolset)) {
  */
 export const writeToolsetComponentDefinition = (
   oldFile: string,
-  toolsets: Toolset[]
+  toolsets: Toolset[],
+  defaultToolset: string | null
 ): string | null => {
   const insertionPoint = findTopLevelPositionBeforeSlashCommand(oldFile);
   if (insertionPoint === null) {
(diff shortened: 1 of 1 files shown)


y0usaf:
Fixed. The fallback is now applied consistently to all four `state.toolset` reads in the file:

- `writeToolFetchingUseMemo` (tool restriction) — already had it
- `writeToolsetComponentDefinition` (\/toolset UI) — now passes `defaultToolset` and applies `?? ${fallback}`
- `insertShiftTabAppStateVar` (status line) — same
- `addCurrentToolsetAtToolChangeComponentScope` (mode-change component) — same

All four generated selectors will now fall back to the configured default toolset on first render with legacy persisted state.

PR #593, opened 2026-03-07
feat: Add allow-custom-agent-models patch

GMNGeoffrey:
Claude Code validates the `model` field in custom agent frontmatter (~/.claude/agents/*.md) against a hardcoded list of aliases (sonnet, opus, haiku, etc.). This prevents using arbitrary model names like "gemini-2.5-flash" or "oai@gemini-3-pro-preview".

This patch removes that restriction so that arbitrary model names pass through to the API. On its own this would cause API errors against the real Anthropic API, but it enables a useful workflow: running a local proxy (e.g. Claudish) that intercepts requests and routes non-Claude model names to appropriate backends with format translation.

Two locations are patched:
1. The Zod schema in UDD that validates against enum(<list variable>) is relaxed to accept any string.
2. The validation flag that gates whether the model field is included in the returned agent definition is changed to accept any non-empty string rather than only specific aliases. The search for this validation uses the captured list variable name from the Zod schema for robustness (the variable name has already changed across the few CC versions I tested).

Verified against real binaries: 2.1.58, 2.1.62, 2.1.69, 2.1.70.

The patch is opt-in (Features group, default: false). Enable with
`allowCustomAgentModels: true` in tweakcc config.

## Summary by CodeRabbit

* **New Features**
  * Added an "Allow custom agent models" setting with a toggle in the Misc/Settings UI to permit arbitrary model names in custom agent frontmatter.
* **Tests**
  * Added test coverage validating the transformation and handling of various model-validation patterns.
* **Integrations**
  * Hooked the new setting into the patching flow so the permissive model handling is applied when enabled.

### DIFF
--- a/data/prompts/prompts-2.1.70.json
+++ b/data/prompts/prompts-2.1.70.json
@@ -1,0 +1,3620 @@
+{
+  "version": "2.1.70",
+  "prompts": [
+    {
+      "name": "Agent Prompt: Agent creation architect",
+      "id": "agent-prompt-agent-creation-architect",
+      "description": "System prompt for creating custom AI agents with detailed specifications",
+      "pieces": [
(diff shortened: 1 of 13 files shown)


georpar:
Great feature; thank you!

PR #655, opened 2026-04-01
Fix: Bun bytecode extraction — fall back to npm source for patchable JS

LeonFedotov:
## Summary

Fixes patch failures on native binary installs where the extracted JS contains Bun bytecode (`// @bun @bytecode` prefix). Function bodies in bytecode are compiled and can't be regex-patched, causing most tweakcc patches to silently fail.

**Fix**: When bytecode is detected after extraction, fetch the matching version's readable `cli.js` from the npm registry. When repacking, clear the bytecode field so Bun uses the patched source JS.

Fixes #628, #629, #635, #645, #651.

## How it works

```
Native binary extraction (before this fix):
  LIEF extract → module.contents = "// @bun @bytecode..." → regex fails ✗

Native binary extraction (after this fix):
  LIEF extract → detect bytecode prefix → npm pack @anthropic-ai/claude-code@X.Y.Z →
  extract cli.js (readable) → patch regex succeeds ✓ →
  repack with bytecode field cleared → Bun runs patched source JS
```

### Why bytecode appears

Recent Claude Code native binaries (2.1.81+) are compiled with Bun's bytecode format. The module graph stores:
- `contents`: source wrapped with `// @bun @bytecode` prefix — function bodies are compiled bytecode, not readable JS
- `bytecode`: the separate compiled bytecode blob

When repacking with patched source, the stale `bytecode` field must be cleared (zeroed) so Bun falls back to interpreting the source JS instead of executing the outdated bytecode.

## Changes

| File | Change |
|---|---|
| `src/nativeInstallation.ts` | `fetchNpmSource()` — downloads cli.js from npm via `npm pack`; bytecode detection in `extractClaudeJsFromNativeInstallation`; clear bytecode on repack |
| `src/nativeInstallationLoader.ts` | Pass `version` parameter through to extraction |
| `src/patches/index.ts` | Pass `ccInstInfo.version` to extraction call |

## Edge cases

- **No network**: Falls back to returning the bytecode content as-is (same behavior as before this fix)
- **Version not on npm**: Same fallback — returns bytecode content
- **npm installs**: Unaffected — they already have readable `cli.js`
- **Version detection callers** (`installationDetection.ts`, `lib/content.ts`): Don't pass version, so no npm fetch is triggered — they only need the version string from comments, which is readable even in bytecode
- **npm pack latency**: ~3-5s on first run (cached by npm afterward)

## Testing

- [x] `pnpm lint` passes (typecheck + ESLint)
- [x] `pnpm run test` passes (221 total, 0 failures)
- [x] `pnpm prettier --check src` passes
- [ ] Manual test: native binary 2.1.89 with bytecode → npm fallback → patches apply

## Summary by CodeRabbit

* **Bug Fixes**
  * Detects non-readable native bytecode and falls back to fetching readable package source when extraction fails.

* **New Features**
  * Option to clear embedded bytecode when repacking native installers for a readable output.
  * Read/write content operations now convey and accept a bytecode-clear flag so patching respects native bytecode state.

* **Tests**
  * Updated and added tests/mocks to cover the new return shape and repack behavior.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -400,7 +400,7 @@ export async function handleUnpack(
     `Extracting JS from native binary: ${chalk.cyan(installation.path)} (v${installation.version})`
   );
 
-  const content = await readContent(installation);
+  const { content } = await readContent(installation);
 
   await fs.writeFile(outputJsPath, content, 'utf8');
 
@@ -448,7 +448,7 @@ export async function handleRepack(
 
   const newJs = await fs.readFile(inputJsPath, 'utf8');
 
-  await writeContent(installation, newJs);
+  await writeContent(installation, newJs, false);
 
   console.log(
     chalk.green(
@@ -471,7 +471,7 @@ async function handleAdhocPatchString(
   installation: Installation,
   skipConfirmation = false
 ): Promise<void> {
-  const content = await readContent(installation);
+  const { content, clearBytecode } = await readContent(installation);
 
   let modified: string;
   let count: number;
@@ -531,7 +531,7 @@ async function handleAdhocPatchString(
     return;
   }
 
-  await writeContent(installation, modified);
+  await writeContent(installation, modified, clearBytecode);
 
   console.log(
     chalk.green(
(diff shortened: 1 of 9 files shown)


LeonFedotov:
@coderabbitai review

darthShadow:
FWIW, `2.1.113` onwards has removed the `cli.js` and replaced it with a native bun executable: https://github.com/anthropics/claude-code/blob/main/CHANGELOG.md#21113

PR #658, opened 2026-04-02
fix: bypass keybinding customization gate for nonessential traffic mode

VitalyOstanin:
## Summary

Custom keybindings from `~/.claude/keybindings.json` are silently ignored when `CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC=1`. The `tengu_keybinding_customization_release` feature flag defaults to `false` without GrowthBook, causing `isKeybindingCustomizationEnabled()` to block all user keybinding loading.

This patch force-enables the gate function (`return !0;` before the GrowthBook check), same technique as worktreeMode, voiceMode, and channelsMode patches.

## Changes

- `keybindingCustomization.ts` — patch implementation with idempotency check
- `keybindingCustomization.test.ts` — tests for success, already-patched, and pattern-not-found
- `index.ts` — register patch as `ALWAYS_APPLIED` (no config condition)

## Test plan

- [x] `pnpm test` — 224 tests pass
- [x] `pnpm lint` — no errors
- [x] Applied to CC 2.1.89 native binary — patch succeeds
- [x] Verified custom keybindings load with `CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC=1`

## Summary by CodeRabbit

* **New Features**
  * Added a keybinding customization patch that forces custom keybindings to be enabled when a specific environment variable is set, and it runs through the standard patch flow.

* **Tests**
  * Added a dedicated test suite to verify behavior for already-patched files, unpatched files that match the expected pattern, and files that don’t match (ensuring a safe null result).

### DIFF
--- a/src/patches/index.ts
+++ b/src/patches/index.ts
@@ -77,6 +77,7 @@ import { writeVoiceMode } from './voiceMode';
 import { writeChannelsMode } from './channelsMode';
 import { writeClearScreen } from './clearScreen';
 import { writeSessionColor } from './sessionColor';
+import { writeKeybindingCustomization } from './keybindingCustomization';
 import {
   restoreNativeBinaryFromBackup,
   restoreClijsFromBackup,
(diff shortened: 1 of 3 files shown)


bl-ue:
@VitalyOstanin I'm sorry for the delay on reviewing this—would you mind rebasing/merging main and testing that it works on CC 2.1.195, assuming you still think it's relevant?

VitalyOstanin:
@bl-ue Thanks! I've rebased onto the latest main (conflicts are now resolved, the PR shows MERGEABLE) and verified the patch on CC 2.1.195: it applies cleanly to the native binary and the patched binary launches fine.

Full disclosure: I no longer need this patch myself, so I don't have a strong stake in it landing. It still works and the feature flag gate is still present in 2.1.195, so it's functionally relevant — but I'll leave it entirely up to you whether to merge or close it. No hard feelings either way.

bl-ue:
@VitalyOstanin Thank you!  Very cool!  If it works, I'm for merging it.  Thank you also for the fast responses!

PR #713, opened 2026-05-01
fix(patchesAppliedIndication): defer findBoxComponent lookup to PATCH 3

risenowrise:
## Problem

`writePatchesAppliedIndication` looks up `findBoxComponent` unconditionally near the top, then aborts the entire indicator if Box isn't found. But only PATCH 3 (the patches-applied list block) actually uses the Box component — PATCH 1 (`--version` string) and PATCH 2 (header banner `+ tweakcc v…` tag) don't reference it at all.

This bites on CC 2.1.126, where the Box function is now:

```js
function o21({children:H,ref:_,tabIndex:q,autoFocus:K,onClick:O,onFocus:T, ...})
```

None of `findBoxComponent`'s four strategies match this shape (Method 2 expects `flexWrap` as the second destructured key; Method 4 expects a single-arg `function NAME(A){let q=memo(N)…}`). Result: the whole patch fails with `failed to find Box component` and CC users on 2.1.126 lose both the version-output tweak and the header banner indicator that don't depend on Box at all.

## Fix

Move the `findBoxComponent` call inside the `if (showPatchesApplied)` branch and `return content` (preserving PATCH 1 + PATCH 2) when Box can't be located, instead of returning `null` and discarding everything. PATCH 3 is the one feature that genuinely needs Box detection — it should be the only one that gets skipped when detection lags.

## Verified on

CC 2.1.126 — both ` --version` output and TUI header banner now show the tweakcc tag (PATCH 1 + PATCH 2 active). PATCH 3 logs `PATCH 3 skipped (Box component not located on this CC version)` and continues. Once `findBoxComponent` is updated for 2.1.126's Box shape, PATCH 3 will start applying again with no further changes here.

## Note

This is a minimal, targeted fix — it doesn't try to fix `findBoxComponent` itself for 2.1.126. That's a separate change.

---

Authored with help from Claude Opus 4.7 (claude-opus-4-7) via Claude Code.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved patch application resilience and robustness by restructuring component checks. The patching system now preserves all successfully applied modifications even when later patch components are unavailable, ensuring better continuity and preventing loss of earlier valid changes. This delivers a more reliable patching experience in various system configurations.

### DIFF
--- a/src/patches/patchesAppliedIndication.ts
+++ b/src/patches/patchesAppliedIndication.ts
@@ -475,14 +475,6 @@ export const writePatchesAppliedIndication = (
     return null;
   }
 
-  const boxComponent = findBoxComponent(fileContents);
-  if (!boxComponent) {
-    console.error(
-      'patch: patchesAppliedIndication: failed to find Box component'
-    );
-    return null;
-  }
-
   // PATCH 2: Add tweakcc version to all header paths.
   // Path A: SyK banner borderText (chalk template literal)
   // Path B: SyK compact borderText (chalk call)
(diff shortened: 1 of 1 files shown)


bl-ue (approved):
Thank you!!

PR #714, opened 2026-05-03
fix(settings): widen theme schema to accept custom theme IDs

### LINKED ISSUE #702: Custom theme selection writes invalid theme ID to settings.json, causing entire config to be skipped
## Problem

When a custom tweakcc theme (e.g. `winter`) is selected via `/theme` in CC, CC writes `"theme": "winter"` to `settings.json`. On next startup, CC validates `settings.json` against its internal schema, which only accepts a fixed enum of built-in theme IDs:

```
["dark", "light", "light-daltonized", "dark-daltonized", "light-ansi", "dark-ansi"]
```

Because `"winter"` is not in this list, CC rejects the **entire `settings.json` file** (not just the `theme` field):

```
Settings Error
C:\Users\...\settings.json
 └ theme: Invalid input

Files with errors are skipped entirely, not just the invalid settings.
```

This causes all other settings — including `statusLine`, hooks, plugins — to stop working until the user manually fixes the file.

## Steps to reproduce

1. Apply tweakcc with a custom theme (e.g. `winter`)
2. In CC, run `/theme` and select the custom theme
3. Restart CC — settings error appears, status bar and other customizations are gone

## Expected behavior

Custom theme IDs should be valid. The themes patch should also patch the settings schema validation to include user-defined theme IDs.

## Suggested fix

The relevant arrays in the binary:
```js
K8$ = ["dark", "light", "light-daltonized", "dark-daltonized", "light-ansi", "dark-ansi"]
_8$ = ["auto", ...K8$]
```

The themes patch could extend these arrays with the user's custom theme IDs.

zeshuochen:
## Problem

When a user configures custom tweakcc themes and selects one (e.g. `winter`), CC writes the bare theme ID to `settings.json`. However, CC's Zod schema for the `theme` field only accepts:

1. Built-in theme IDs (`auto`, `dark`, `light`, `light-daltonized`, `dark-daltonized`, `light-ansi`, `dark-ansi`)
2. Strings starting with `"custom:"`

A bare ID like `"winter"` fails both checks. The `.catch(void 0)` on the field silently resets it to `undefined`, so the theme is discarded on every startup. Only a legacy fallback via `~/.claude.json` keeps it working — a fragile dependency that may not always be present.

Fixes #702.

## Solution

Add a new patch `settingsTheme` that replaces:

```js
// Before
theme:VAR.union([VAR.enum(THEMES),VAR.string().startsWith("custom:").transform(fn)])
         .optional().catch(void 0)

// After  
theme:VAR.string().optional().catch(void 0)
```

This mirrors the approach used in `allowCustomAgentModels` (replacing a narrow enum with `z.string()`). The patch applies to both the user settings and managed settings schema instances. It is gated behind the same condition as the `themes` patch — only active when custom themes are configured.

## CC version tested

Verified against `claude 2.1.126`: the regex matches both schema instances (user settings + managed settings) and correctly transforms them.

## Test plan

- [x] TypeScript compiles without errors (`tsc --noEmit`)
- [x] All 221 existing tests pass
- [x] Regex verified against `claude 2.1.126` binary — matches 2 occurrences
- [ ] Install tweakcc with a custom theme → select it → restart CC → verify theme persists in `settings.json`

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a settings theme patch that improves validation and handling of user theme configurations; it now conditionally applies when custom themes are present and differ from defaults, reducing incorrect theme application and ensuring safer migration of theme settings.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,22 +1,24 @@
 {
   "name": "tweakcc",
-  "version": "3.4.0",
+  "version": "4.0.11",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "tweakcc",
-      "version": "3.4.0",
+      "version": "4.0.11",
       "license": "MIT",
       "dependencies": {
         "chalk": "^5.5.0",
+        "cli-spinners": "^3.4.0",
         "commander": "^14.0.0",
         "diff": "^8.0.3",
         "globby": "^14.1.0",
         "gray-matter": "^4.0.3",
         "ink": "^6.1.0",
         "ink-link": "^4.1.0",
-        "node-lief": "^1.0.0",
+        "node-lief": "^1.1.1",
+        "oxfmt": "^0.28.0",
         "react": "^19.1.1",
         "wasmagic": "^1.0.7",
         "which": "^6.0.0"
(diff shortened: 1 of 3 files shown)


bl-ue:
Hi @zeshuochen, thank you for the PR!  I think a simpler implementation would be to merge this functionality into the existing theme-related patching instead of creating a new patch, so I merged #773 instead.  We appreciate the contribution, however!

Issue #756, opened 2026-05-28
8 patches failing on Claude Code 2.1.146+ (last successful prompts: 2.1.138)


rpickmans:
I've been tracking patch-application status across 6 consecutive Claude Code releases (2.1.146 → 2.1.150 → 2.1.152 → 2.1.153) and the same 8 patches have failed on every one of them. tweakcc is at the latest v4.0.13 (released 2026-05-10), and based on `npx tweakcc@latest --apply` output, the most recent CC prompt set shipped in tweakcc is for 2.1.138 — no prompts have been added for any 2.1.146+ release.

Reproduce: `npx -y tweakcc@latest --apply` against Claude Code 2.1.153 yields:

**Always-Applied (4 failing):**
- ✗ Verbose property
- ✗ Opusplan[1m] support
- ✗ Fix LSP support
- ✗ Statusline update throttling correction

**Misc (1 failing):**
- ✗ Thinker format

**Features (3 failing):**
- ✗ Toolsets
- ✗ MCP batch size
- ✗ User message display

The other 17 patches/data entries apply cleanly (✓).

Is the 2.1.138-baseline prompt set still considered current, or are prompts for 2.1.146+ in progress? If the latter, happy to test pre-release patches against my installed CC 2.1.153.

Environment: macOS Darwin 25.5.0 (arm64), CC 2.1.153, tweakcc 4.0.13.

bl-ue:
Hi @rpickmans!  We're aware of the patching failures and we're working to fix them.

PR #760, opened 2026-05-30
Prompts for 2.1.158

mike1858:


rpickmans:
Thanks for putting this up. I'm on CC 2.1.158 (macOS, arm64) - the build this PR targets - so I can run a patch pass against it once this merges and report back. This is the same setup as #756, where the same 8 patches have failed on every release since 2.1.146. If it's useful, I can report which of those start applying again under the 2.1.158 prompt set and which still fail; I suspect the prompt-data lag and the native-binary failures (the objArrMatch / Bun repack errors) are two separate problems, and that only the former may be resolved here. Glad to test against the exact version if that would help.

rpickmans:
Ran the patch pass against CC 2.1.158 now that this merged. The system-prompt customizations applied cleanly, so #756's prompt layer is fixed here - and since I'm still on the published 4.0.13, that confirms the prompts are fetched at runtime rather than bundled into the release.

The 8 patches from #756 still fail, unchanged: Verbose property, Opusplan[1m], Fix LSP, and Statusline throttling (Always Applied); Thinker format (Misc); Toolsets, MCP batch size, and User message display (Features). These read as a separate problem from the prompt-data lag - the errors look like anchor/objArrMatch misses against the native binary rather than missing prompt data, so I don't expect more prompt PRs to move them. No corruption, though; `claude --version` is fine and the patches-applied indicator shows.

Glad to test any native-binary anchor fixes against 2.1.158 if that would help.

PR #808, opened 2026-06-17
Fix system prompt application on Claude Code 2.1.179: match Latin-1 \xHH escapes

lukehutch:
## Problem

On Claude Code **2.1.179**, `tweakcc --apply` fails to apply ~10 system prompts, each logged as:

```
Could not find system prompt "Skill: /catch-up periodic heartbeat" in cli.js (using regex ...)
```

The affected prompts all contain a **Latin-1 character** (U+0080–U+00FF) — most commonly `·` (middle dot, in the `<title> · <time> · <attendees>` notification format), but also `×`, `°`, etc.

## Root cause

`escapeNonAsciiForRegex` (in `src/systemPromptSync.ts`) builds the cli.js search regex by matching each non-ASCII char as **either the literal char or its `\uXXXX` escape**:

```js
const escaped = `\\\\u${codePoint.toString(16).padStart(4, '0')}`;
return `(?:${literal}|${escaped})`;
```

But recent Bun-compiled Claude Code bundles store **Latin-1 chars as 2-digit `\xHH` escapes**, not `\uXXXX` — e.g. `·` is emitted as `\xB7`, `×` as `\xD7`. Because the regex only offered the `\uXXXX` form, any prompt containing such a char failed to match and the whole prompt was skipped.

Verified directly against a 2.1.179 `unpack`: the middle dot appears only as `\xB7` (zero occurrences of `·` or a literal `·`).

This is independent of the `patches-applied-indication` UI abort that `main` already handles; both had to be resolved for `--apply` to fully succeed on 2.1.179.

## Fix

Add the `\xHH` alternative for codepoints `0x80`–`0xFF`. The search regex's existing `i` flag already covers upper/lowercase hex (`\xb7` vs `\xB7`):

```js
const alts = [literal, `\\\\u${codePoint.toString(16).padStart(4, '0')}`];
if (codePoint >= 0x80 && codePoint <= 0xff) {
  alts.push(`\\\\x${codePoint.toString(16).padStart(2, '0')}`);
}
return `(?:${alts.join('|')})`;
```

## Result

On a real Claude Code 2.1.179 install, this takes `--apply` from **10 "could not find" → 0**: every system prompt now locates and applies.

## Tests

- Exported `buildSearchRegexFromPieces` and added regression tests asserting a piece containing `·`/`×` matches the `\xHH`, `\uXXXX`, **and** literal forms — and that non-Latin-1 chars (e.g. `—`) still match via `\uXXXX`/literal.
- `pnpm lint` and `pnpm test` pass (277 tests).

Complements #785 (Text component detection for 2.1.170) and #798 (opusplan1m for 2.1.175+).

## Release Notes

* **Bug Fixes**
  * Enhanced Unicode and UTF-8 escape handling to ensure prompts are correctly recognized regardless of how special characters are encoded by upstream bundles, preventing prompt skipping issues.

* **Tests**
  * Added comprehensive test coverage for Unicode character escape variants.

### DIFF
--- a/src/systemPromptSync.ts
+++ b/src/systemPromptSync.ts
@@ -1048,9 +1048,18 @@ const escapeNonAsciiForRegex = (text: string): string => {
   // eslint-disable-next-line no-control-regex
   return text.replace(/[^\x00-\x7F]/g, char => {
     const codePoint = char.charCodeAt(0);
-    const escaped = `\\\\u${codePoint.toString(16).padStart(4, '0')}`;
-    // Match either the literal character OR the escaped version
-    return `(?:${char.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')}|${escaped})`;
+    const literal = char.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+    // Match the literal char OR its escaped forms. Minified bundles store BMP
+    // non-ASCII as \uXXXX, but Latin-1 chars (0x80–0xFF) are emitted as \xHH
+    // (2-digit hex, e.g. "·" -> \xB7, "×" -> \xD7). The search regex's 'i' flag
+    // handles upper/lowercase hex. Without the \xHH alternative, any prompt
+    // containing such a character (·, ×, °, é, …) fails to match on recent
+    // Claude Code builds, so the whole prompt is skipped.
+    const alts = [literal, `\\\\u${codePoint.toString(16).padStart(4, '0')}`];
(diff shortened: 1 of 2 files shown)


lukehutch:
Update: also verified on **Claude Code 2.1.181** (released after this PR was opened) — `--apply` lands every system prompt with **0 "could not find"** there too. The bug is general (any build that emits Latin-1 chars like `·`/`×` as `\xHH`), not specific to one CC version.

bl-ue (approved):
Thank you!

PR #819, opened 2026-06-26
Fix crash when a system prompt regex is too complex to compile (#753)

### LINKED ISSUE #753: Large system-prompt patches crash with 'Invalid regular expression … Stack overflow' (Migration Guide, Building LLM apps, etc.)
## Symptom

`tweakcc --apply` crashes with:

```
SyntaxError: Invalid regular expression: /# Model Migration Guide(?:\n|\n)(?:\n|\n)> ...[truncated, ~70KB pattern]/gis: Stack overflow
    at RegExp.exec (<anonymous>)
    at Pt (file:///.../dist/config-...mjs:168:5048)
    at async ci (...)
    at async xn (...)
```

Exit code 1. The crash aborts `--apply` partway through, so **no patches land at all** (`changesApplied: false` in config), even patches that succeeded earlier in the run.

The culprit prompts are the large ones (Migration Guide ~70 KB, Building LLM-Powered Applications with Claude ~similar). The regex tweakcc builds from the markdown body is large enough to overflow V8's regex engine stack on both `new RegExp(pattern, 'sig')` construction AND subsequent `regex.exec(content)` calls.

## Repro

1. Cold-install tweakcc (or any path where these prompts get seeded into `~/.tweakcc/system-prompts/`)
2. `node dist/index.mjs --apply` against any CC version that has these prompts in `data/prompts/prompts-X.Y.Z.json`
3. Crashes on the first large prompt

## Root cause

In `src/patches/systemPrompts.ts` around line 136 (and the dist-minified equivalent):

```ts
const globalPattern = new RegExp(regex, 'sig');
// ...
while ((mm = globalPattern.exec(content)) !== null) { ... }
```

Neither `new RegExp(...)` construction nor `.exec()` is wrapped in try/catch. For prompts whose markdown body is large (tens of KB) and contains many escape-alternation constructs like `(?:\n|\n)`, `(?:'|\\')`, `(?:—|\u2014)`, the resulting compiled pattern has enough alternation depth that V8 hits its stack limit.

## Suggested fix

Wrap both the constructor and the exec loop in try/catch, skip the offending prompt with a warning, and continue:

```ts
let globalPattern: RegExp;
let simplePattern: RegExp;
try {
  simplePattern = new RegExp(regex, 'si');
  globalPattern = new RegExp(regex, 'sig');
} catch (err) {
  console.log(chalk.yellow(
    \`Skipped regex too complex for \"\${prompt.name}\": \${(err as Error).message}\`
  ));
  results.push({
    id: promptId, name: prompt.name, group: PatchGroup.SYSTEM_PROMPTS,
    applied: false, details: 'regex too complex',
  });
  continue;
}

const allMatches: RegExpExecArray[] = [];
try {
  let mm: RegExpExecArray | null;
  while ((mm = globalPattern.exec(content)) !== null) {
    allMatches.push(mm);
    if (mm[0].length === 0) globalPattern.lastIndex++;
  }
} catch (err) {
  console.log(chalk.yellow(
    \`Skipped exec too complex for \"\${prompt.name}\": \${(err as Error).message}\`
  ));
  results.push({
    id: promptId, name: prompt.name, group: PatchGroup.SYSTEM_PROMPTS,
    applied: false, details: 'exec too complex',
  });
  continue;
}
```

Locally validated this in the skrabe/tweakcc-fixed fork's dist — after wrapping, `--apply` completes with `changesApplied: true`, ~80 patches land, and the offending Migration Guide + Building-LLM-apps prompts emit `Skipped regex too complex` warnings.

## Workaround

Either:
- Truncate the offending markdown bodies in `~/.tweakcc/system-prompts/` to a short stub (regex shrinks, no stack overflow). Per-prompt; each large prompt needs trimming.
- Apply the local try/catch patch to dist (clobbered on every `pnpm build`).

## Environment

- OS: Windows 10 Home
- Node: v22.17.0
- CC versions tested: 2.1.145 (`~/.local/bin/claude.exe`), 2.1.149 (Desktop bundle)
- tweakcc / tweakcc-fixed commit: 9b12ab2 (skrabe fork; this code path is unchanged from upstream)

## Related

Companion issue #752 (`tokenCountRounding` corrupts CC 2.1.149). Same broader pattern: tweakcc patches against the 15 MB minified bundle need defensive guards against (a) CC minifier shape drift and (b) regex stack/backtracking explosions on large prompt content.

StreamDemon:
## Summary

On Node <=22, `applySystemPrompts` crashes with a regex stack overflow when a
system prompt's generated search pattern is large enough — e.g. the Model
Migration Guide (~138 KB pattern) or Building LLM-powered applications (~80 KB).
`new RegExp(regex, 'si')` is unguarded, so the throw aborts the entire `--apply`
and **no patches land** (`changesApplied: false`), even ones that already
succeeded earlier in the run. Reported in #753.

This implements the defensive guard suggested by @arikyeo in that issue.

## Changes

- Wrap the per-prompt regex construction and `content.match` in a try/catch in
  `applySystemPrompts`. On failure: log a warning, push a `PatchResult` with
  `applied: false, details: 'regex too complex'`, and `continue` to the next
  prompt instead of letting the throw abort the run.
- No behavior change for any prompt whose regex compiles.

## Testing

- [x] Tests added/updated — `src/patches/systemPrompts.test.ts`: one test proves an
  uncompilable pattern is skipped (not thrown); another proves a later prompt still
  applies after an earlier one fails.
- [x] All existing tests pass — full `vitest run`: 276 passed, 5 skipped. `tsc --noEmit`,
  `eslint src`, `prettier --check`, and `build:dev` all clean.
- [ ] The Node <=22 stack-overflow itself is runtime-specific (validated per #753 on
  Node 22); the guard is covered deterministically on Node 24 via an invalid-pattern test.

## Related Issues

Closes #753

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a crash that could occur when a system prompt uses a search pattern too complex to compile.
  * `--apply` now skips the problematic prompt and continues processing the rest instead of stopping early.
  * Improved reporting so skipped prompts are clearly marked as not applied, while valid prompts still apply normally.

* **Documentation**
  * Updated the changelog with the latest crash fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- Fix crash when a system prompt's search regex is too complex to compile, which aborted the entire `--apply` on Node <=22 (#753) - @StreamDemon
+
 ## [v4.1.1](https://github.com/Piebald-AI/tweakcc/releases/tag/v4.1.1) - 2026-06-17
 
 - Fix system prompt application on Claude Code 2.1.179: match Latin-1 \xHH escapes (#808) - @lukehutch
--- a/src/patches/systemPrompts.test.ts
+++ b/src/patches/systemPrompts.test.ts
@@ -474,5 +474,61 @@ describe('systemPrompts.ts', () => {
       expect(result.results[0].skipped).toBe(true);
       expect(result.results[0].applied).toBe(false);
     });
+
+    it('should skip a prompt whose regex fails to compile instead of throwing', async () => {
+      const mockPromptData = buildMockPromptData({
+        promptId: 'uncompilable-prompt',
+        prompt: { name: 'Uncompilable Prompt', content: 'unused' },
+        regex: '(', // invalid pattern: new RegExp('(', 'si') throws (unterminated group)
+        getInterpolatedContent: () => 'unused',
+        pieces: ['unused'],
+      });
+
+      setupMocks(mockPromptData);
+
+      const cliContent = 'desc:"some content"';
+
+      const result = await applySystemPrompts(cliContent, '1.0.0', false);
+
+      expect(result.newContent).toBe(cliContent);
+      expect(result.results).toHaveLength(1);
+      expect(result.results[0].applied).toBe(false);
+      expect(result.results[0].details).toContain('too complex');
+    });
+
+    it('should continue applying remaining prompts after one regex fails to compile', async () => {
+      const badPrompt = buildMockPromptData({
+        promptId: 'bad-prompt',
+        prompt: { name: 'Bad Prompt', content: 'unused' },
+        regex: '(',
+        getInterpolatedContent: () => 'unused',
+        pieces: ['unused'],
+      });
+      const goodPrompt = buildMockPromptData({
+        promptId: 'good-prompt',
+        prompt: { name: 'Good Prompt', content: 'New content' },
+        regex: 'Original text',
+        getInterpolatedContent: () => 'New content',
+        pieces: ['Original text'],
+      });
+
+      vi.mocked(promptSync.loadSystemPromptsWithRegex).mockResolvedValue([
+        badPrompt,
+        goodPrompt,
+      ]);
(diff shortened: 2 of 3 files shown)


bl-ue (approved):
Thank you for your contribution!

Issue #822, opened 2026-06-27
Tracking: multiple patches broken on Claude Code 2.1.195 (createElement → JSX automatic-runtime migration)


xRevenantPulse:
Several tweakcc patches stopped matching the Claude Code bundle around the **2.1.x** series. This is a tracking issue for a full audit performed against **CC 2.1.195**.

**How this was verified:** each patch's `writeXxx()` was run against the JS bundle extracted from an installed CC `2.1.195` native binary (via tweakcc's own `extractClaudeJsFromNativeInstallation`), and individual anchors were tested against that bundle. "Broken" = the locating regex/anchor returns no match (patch returns `null` or silently no-ops), or the patch applies at the wrong location via a fallback.

**Root cause (shared by most of these):** CC 2.1.195 compiles its UI with the **React JSX automatic runtime** — the bundle has ~6603 `jsx(` + ~3176 `jsxs(` calls and only ~40 `.createElement(` (all DOM, none from the React var). Anchors written against `X.default.createElement(...)` no longer match.

A handful of patches are **not** affected: `themes` (all anchors fine), most thinker/UI patches, and several that are correctly version-gated off or now redundant because CC supports the behavior natively (`worktree-mode`, `mcp-non-blocking`, `fix-lsp-support`, `allow-custom-agent-models`).

### Broken / partial on 2.1.195 (grouped by area → one PR each)

**Always-on (affect every user):**
- [x] #823 — `verbose-property` (PR #833, merged)
- [x] #824 — `user-message-display` (silent), `patches-applied-indication` (applied-list dead) (PR #834, merged)

**Config-gated — broken when the feature is enabled:**
- [x] #825 — `thinker-symbol-width`, `thinker-format` (PR #835, merged)
- [x] #826 — `input-box-border`, `input-pattern-highlighters` (PR #840, merged)
- [x] #827 — `table-format`, `token-count-rounding`, `context-limit` (PR #839, merged)
- [x] #828 — `suppress-rate-limit-options` (broken), `suppress-line-numbers` (partial), `suppress-native-installer-warning` (partial) (PR #838, merged)
- [x] #829 — `toolsets` (PR #841, merged)
- [x] #830 — `auto-accept-plan-mode` (PR #842, merged)
- [x] #831 — `hide-startup-clawd` (applies at wrong location) (PR #836, merged)
- [x] #832 — `channels-mode` (partial — cosmetic notice only) (PR #837, merged)

[...]

PR #838, opened 2026-06-27
Fix suppress-rate-limit-options, suppress-line-numbers, and suppress-native-installer-warning patches for Claude Code 2.1.195

### LINKED ISSUE #828: Suppress patches broken/partial on CC 2.1.195: rate-limit-options + line-numbers + native-installer-warning
**Measured on 2.1.195:**
- `suppress-rate-limit-options` — `writeSuppressRateLimitOptions()` returns `null` (`failed to find onOpenRateLimitOptions pattern`); the leading `.createElement` anchor no longer matches (JSX runtime). **Fully broken.**
- `suppress-line-numbers` — *partial*: the core line-number suppression applies correctly; only prompt-rewrite **p2** fails because the target text moved into a template literal where the inner backticks around `:` are now backslash-escaped, so the bare `\`:\`` anchor matches 0 times.
- `suppress-native-installer-warning` — *partial*: applies (p4 "Native installation exists but ~/.local/bin not in your PATH" ×2 and p5 the `echo export PATH` fix line match), but **3 of 5** patterns now miss (moved/renamed doctor-diagnostic strings).

**Likely fix:** for rate-limit, allow `(?:\.createElement|\.jsx)` in pattern 1; for line-numbers p2, escape the inner backticks in the regex to match the template-literal form; for native-installer-warning, update the 3 stale diagnostic anchors (and drop any obsolete ones).

Part of #822

StreamDemon:
## Problem

Three "suppress" patches broke on recent Claude Code (2.1.193+/2.1.195):

- **suppress-rate-limit-options** — anchored on `.createElement(...)` call sites. CC migrated to the JSX automatic runtime (`.jsx`/`.jsxs`) and reordered the props, so both patterns missed and the patch returned `null`.
- **suppress-line-numbers** — the second Read-tool prompt rewrite never matched. Its anchor expects a literal `` `:` `` inside the prompt, but that text lives in a template literal, so the inner backticks are escaped in the bundle (`` \`:\` ``). The formatter and the first prompt rewrite still applied; only this line silently no-op'd.
- **suppress-native-installer-warning** — the startup PATH warning is now emitted as `s.push({message:`Native installation exists but …`,userActionRequired:!0,type:"path"})` with separate Windows and Unix branches. The existing string-blanking patterns only blanked the Unix message's first sentence (leaving the rest behind) and missed Windows entirely.

## Fix

- **suppress-rate-limit-options** — replace the two `.createElement`-anchored patterns with a single render-agnostic anchor: `showAllInTranscript:VAR,(agentDefinitions:VAR,)?onOpenRateLimitOptions:(VAR)`. It nulls the callback at both the normal and the virtual-scroll transcript renderers, and the `=!1` default on the props *definition* naturally excludes that site (so the destructuring is never corrupted). This matches both the old `createElement` and the new `jsx` layouts.
- **suppress-line-numbers** — match the escaped inner backticks so the prompt rewrite applies again.
- **suppress-native-installer-warning** — remove the whole startup `push({…,type:"path"})` for both platforms. The legacy npm→native banner pattern is kept for older CC; the 2.1.146–2.1.158 `claude doctor` string-blanking patterns are dropped — they match nothing on current CC and are outside this patch's stated "at startup" scope.

## Verification

Ran each `writeXxx` against freshly extracted **2.1.195** and **2.1.193** bundles:

- suppress-rate-limit-options: `null` → applies; exactly **2** `onOpenRateLimitOptions:()=>{}` injected; props definition untouched.
- suppress-native-installer-warning: startup-push count **2 → 0** on both branches (Windows + Unix).
- suppress-line-numbers: both prompt rewrites now apply.
- `node --check` passes on the fully patched bundle (the native-installer fix changes control flow, not just a string).

Added focused unit tests for all three patches (10 cases). `pnpm lint` and `pnpm test` are green (full suite passing).

Part of #822. Closes #828.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved patching compatibility so suppressions work across more variations of prompt and startup message formatting.
  * Refined removal of native installer warnings to match additional related message text while preserving surrounding content.
  * Updated line-number and rate-limit option suppression to handle alternate syntaxes and safely return when no matches are found.
* **Tests**
  * Added unit tests covering prompt rewriting, native warning stripping, alternate syntax cases, and no-match behavior.
* **Documentation**
  * Updated the changelog with the latest fix entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix verbose-property patch for Claude Code's JSX automatic runtime (#833) - @StreamDemon
 - Fix `userMessageDisplay` and `patchesAppliedIndication` patches for Claude Code's JSX automatic runtime, and make `findBoxComponent` JSX-aware (#834) - @StreamDemon
 - Fix thinker-symbol-width and thinker-format patches for Claude Code's JSX automatic runtime (#835) - @StreamDemon
+- Fix suppress-rate-limit-options, suppress-line-numbers, and suppress-native-installer-warning patches for Claude Code 2.1.195 (#838) - @StreamDemon
 
 ## [v4.2.0](https://github.com/Piebald-AI/tweakcc/releases/tag/v4.2.0) - 2026-06-26
 
--- a/src/patches/suppressLineNumbers.test.ts
+++ b/src/patches/suppressLineNumbers.test.ts
@@ -1,0 +1,36 @@
+import { describe, it, expect } from 'vitest';
+import { writeSuppressLineNumbers } from './suppressLineNumbers';
+
+describe('writeSuppressLineNumbers', () => {
+  const formatter =
+    'function f({content:J,startLine:G}){if(!J)return"";let L=J.split(/\\r?\\n/);return L.map(x=>x).join("")}function g(){}';
+  const p1 =
+    '"- Results are returned using cat -n format, with line numbers starting at 1"';
+  const p2 =
+    '`${oYr}. Each line is the line number, a single separator (a tab or \\`:\\`), then the verbatim file content (including any leading whitespace).`';
+
+  it('neutralizes the formatter and rewrites both read-tool prompt lines', () => {
+    const result = writeSuppressLineNumbers(
+      formatter + ';var X=' + p1 + ',Y=' + p2 + ';'
+    );
+    expect(result).not.toBeNull();
+    expect(result).toContain('if(!J)return"";return J}');
+    expect(result).toContain(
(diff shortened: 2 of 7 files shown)


xRevenantPulse:
Addressed both nitpicks in 794f64d:

- `suppressLineNumbers.test.ts`: added a case feeding the unescaped `` `:` `` separator so the optional-backslash branch of the p2 regex is exercised alongside the escaped template-literal form.
- `suppressRateLimitOptions.test.ts`: added a case with `$`-prefixed minified identifiers (`$ne`, `$vt`, …) to lock in `[$\w]+` matching and guard against a `\w+` regression.

Full suite green (323 passed).

bl-ue (approved):

PR #842, opened 2026-06-29
Fix auto-accept-plan-mode patch for Claude Code's JSX automatic runtime

### LINKED ISSUE #830: auto-accept-plan-mode broken on CC 2.1.195 (injects at the wrong return)
**Measured against the pristine `2.1.195.orig` bundle (and confirmed on `2.1.193`).** *(This corrects an earlier draft of the issue that mis-described the handler and the select's location.)*

The only surviving anchor is the `title:"Ready to code?"` locator. Because of a `lastIndexOf('return ')` fallback the patch *applies anyway* — at a garbage location — so it doesn't error but silently doesn't work (and can corrupt an inner callback).

### What's actually in the bundle

- There is **exactly one** plan-approval select, and it sits **after** the title: `iy.jsx(Sr,{options:N,onChange:Dn,onCancel:Ze,onImagePaste:K,pastedContents:d,onRemoveImage:J})`. The `onImagePaste`/`pastedContents`/`onRemoveImage` props are present because one of the select's options is an embedded text input ("No, keep planning") — it is **not** a separate "prompt-input box", and there is no second select before the title.
- The options come from the `Hhm()` builder, where `value:"yes-accept-edits"` lives **~11,078 chars _before_** the title. That's the options *builder*, not a select call.
- `onChange:Dn`, where `Dn=(Mr)=>void Ie(Mr)`, delegates to the real plan-approval handler `Ie`. So the patch's existing `onChange:…,onCancel` extraction already grabs a **correct** handler — the original "wrong handler" framing was a misdiagnosis.

### The actual defect: wrong injection point

The precise `createElement(...,{color:"planMode",title:"Ready to code?"})` anchor is gone (JSX runtime). The component (`function xEc`) ends with a top-level `…else kr=t[105];return kr}` located ~1,802 chars **after** the title. But `lastIndexOf('return ')` (searching the window *before* the title) lands at title−1059 — `…else it=t[66];return it}`, the tail of an **inner memoized callback**. The injection (`Dn("yes-accept-edits-keep-context");return null;`) therefore ends up there, where `Dn` isn't in scope and the early `return null` corrupts that callback.

### Fix

Keep the handler extraction; replace the injection-point logic with a finder for the component's true top-level `return <var>}` *after* the title:

- **Primary:** the React-Compiler memo tail `else <v>=t[<n>];return <v>}` (a backreference keeps it from matching an unrelated memo slot).
- **Fallback:** a string-aware walk of the enclosing function for its last top-level `return` (covers non-memoized / legacy `createElement` layouts and keeps the existing test green).

Drop the `createElement` anchors and the `lastIndexOf('return ')` fallback for this JSX-runtime layout.

Part of #822

StreamDemon:
## Problem

`auto-accept-plan-mode` is broken on CC 2.1.195 (JSX automatic runtime). The precise `createElement(...)` injection anchor is gone, so the patch's `lastIndexOf('return ')` fallback injected `<handler>(...);return null;` at a **garbage location** — the tail of an inner memoized callback ~1,059 chars *before* the `title:"Ready to code?"` locator — instead of the dialog component's true top-level return ~1,802 chars *after* it. The patch *applied* (no error) but silently did nothing, and could corrupt that inner callback.

### Correcting the issue's diagnosis

Measured against the pristine `2.1.195.orig` bundle (and confirmed on `2.1.193`), #830's original "wrong handler / two selects" framing is inaccurate:

- There is **exactly one** plan-approval select, and it sits **after** the title: `iy.jsx(Sr,{options:N,onChange:Dn,onCancel:Ze,onImagePaste:K,…})`. The `onImagePaste`/`pastedContents`/`onRemoveImage` props are present because one of its options is an embedded text input ("No, keep planning") — not a separate "prompt-input box".
- `value:"yes-accept-edits"` lives ~11,078 chars *before* the title, inside the `Hhm()` options *builder* — not a second select.
- `onChange:Dn`, where `Dn=(Mr)=>void Ie(Mr)`, already delegates to the real handler, so the existing `onChange:…,onCancel` extraction was **already correct**. The only real defect is the injection point.

(#830 has been updated to reflect this.)

## Fix

Keep the handler extraction and the prompt/permission rewrites (all their anchors still match on 2.1.195 + 2.1.193). Replace only the injection-point logic:

- **Primary:** the React-Compiler memoized tail `else <v>=t[<n>];return <v>}` — a backreference keeps it from matching an unrelated memo slot. Lands at `else kr=t[105];return kr}` (2.1.195) and `else Ir=t[105];return Ir}` (2.1.193).
- **Fallback:** a string-aware walk of the enclosing function for its last top-level `return` (covers non-memoized / legacy `createElement` layouts and keeps the existing test green).

This drops the `createElement` anchors and the `lastIndexOf('return ')` fallback. Calling the dispatch during render (`<handler>(...);return null;`) is the patch's pre-existing mechanism, restored at the correct location.

## Verification

- Ran the real `writeAutoAcceptPlanMode` against the extracted 2.1.195 (`.orig`) and 2.1.193 bundles: the injection lands at the component's top-level return (`else kr=t[105];Dn("yes-accept-edits-keep-context");return null;return kr}`, and the `wn`/`Ir` equivalent on 2.1.193); the permission anchors (`permission_exit_plan_mode_v2` default → allow, `checkPermissions` → allow) and prompt rewrites still apply; and `node --check` passes on the full patched bundle for both versions.
- Tests: kept the `createElement` fallback case; added a JSX-runtime memoized fixture (asserts injection at the top-level return, not the inner callback before the title) and an idempotency check. `pnpm run lint` + `pnpm run test` green.

Part of #822. Closes #830.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the plan-mode auto-accept patch for JSX automatic runtime output, ensuring the acceptance step is inserted at the correct top-level return (including memoized layouts).
  * Avoids injecting near nested returns.
  * Maintains idempotency when the patch is applied repeatedly.
* **Tests**
  * Added new test cases for JSX-runtime scenarios and verified that re-running the patch produces identical results.
* **Documentation**
  * Updated the changelog with the latest auto-accept-plan-mode fix entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix thinker-symbol-width and thinker-format patches for Claude Code's JSX automatic runtime (#835) - @StreamDemon
 - Fix hide-startup-clawd patch for Claude Code's JSX automatic runtime (#836) - @StreamDemon
 - Fix suppress-rate-limit-options, suppress-line-numbers, and suppress-native-installer-warning patches for Claude Code 2.1.195 (#838) - @StreamDemon
+- Fix auto-accept-plan-mode patch for Claude Code's JSX automatic runtime (#842) - @StreamDemon
 
 ## [v4.2.0](https://github.com/Piebald-AI/tweakcc/releases/tag/v4.2.0) - 2026-06-26
 
--- a/src/patches/autoAcceptPlanMode.test.ts
+++ b/src/patches/autoAcceptPlanMode.test.ts
@@ -17,4 +17,40 @@ describe('writeAutoAcceptPlanMode', () => {
       'h("yes-accept-edits-keep-context");return null;return R.default.createElement'
     );
   });
+
+  it('injects before the component return for the JSX-runtime memoized layout', () => {
+    // Mirrors CC 2.1.195: a single plan-approval select after the title (its
+    // onImagePaste/pastedContents/onRemoveImage props come from an embedded text
+    // input), an inner memoized callback whose return sits BEFORE the title, and the
+    // component's true top-level tail `else kr=t[105];return kr}` AFTER the title.
+    const input =
+      'function xEc(e){let t=Mc.c(106);' +
+      'let it;if(t[66]!==e)it=(q)=>{if(q)return q;return 0},t[66]=e,t[67]=it;else it=t[66];' +
+      'let vt;if(t[80]!==e)vt=R.jsx(Lf,{color:"planMode",title:"Ready to code?",children:xt}),t[80]=e,t[82]=vt;else vt=t[82];' +
+      'let Dn;if(t[88]!==Ie)Dn=(Mr)=>void Ie(Mr),t[88]=Ie,t[89]=Dn;else Dn=t[89];' +
+      'let nn;if(t[90]!==Ze)nn=R.jsx(Sr,{options:N,onChange:Dn,onCancel:Ze,onImagePaste:K,pastedContents:d,onRemoveImage:J}),t[90]=Ze,t[95]=nn;else nn=t[95];' +
+      'let kr;if(t[102]!==tt)kr=R.jsxs(U,{flexDirection:"column",tabIndex:0,autoFocus:!0,onKeyDown:tt,children:[vt,nn]}),t[102]=tt,t[105]=kr;else kr=t[105];return kr}';
+
+    const result = writeAutoAcceptPlanMode(input);
+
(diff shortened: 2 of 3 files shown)


xRevenantPulse:
Thanks for the review! Applied the one nitpick in bf5734a — the memo-tail search now reuses the existing `afterReady` slice instead of re-slicing the identical `readyIdx, readyIdx + 3000` window. Pure refactor (`afterReady.match(...).index` is relative to `readyIdx`, same as before); unit tests and the real-bundle verification on 2.1.195 + 2.1.193 re-confirmed.

PR #855, opened 2026-07-03
Fix input-chevron-color anchor for CC 2.1.199 (isScreenReader field + 3-term guard)

### LINKED ISSUE #846: bug/regression: Input chevron color
tweakcc:4.3.0
cc cli:2.1.185

  Misc Configurable:
    ✓ Model customizations — Access all Claude models with /model, not just latest 3
    ✓ Show more items in select menus — Show 25 items in select menus instead of default 5
    ✓ Patches applied indication — Show "tweakcc patches applied" and tweakcc version inside CC
    ✓ Table format — Tables generated by Claude will be rendered more compactly
    ✗ Input chevron color
    ✓ Thinking block visibility — Thinking blocks outputted by the model will show without Ctrl+O
    ✓ Token count rounding — Round displayed token counts to the nearest multiple of chosen value
    ✓ AGENTS.md (and others) — Support AGENTS.md and others in addition to CLAUDE.md

  Features:
    ✓ Session memory — Enable session memory (auto-extraction + past session search)
    ✓ User message display — User messages in the chat history will be styled

Customizations applied with some failures.
These patching errors do not affect your system prompt patches.
Your system prompt customizations were still applied successfully.
Please open an issue on https://github.com/Piebald-AI/tweakcc/issues/new reporting these patching errors.
Run with --restore/--revert to revert Claude Code to its original state.
FAILED.

no custom color. in config.json

StreamDemon:
Fixes #846.

## Problem
`input-chevron-color` reports ✗ on recent CC (confirmed on **2.1.199**, native binary, Linux x64). The chevron idle color isn't applied.

## Root cause
CC 2.1.199 restructured the chevron component. The patch's anchor expected:

```js
{isLoading:X,themeColor:Y}=Z,W=Y??void 0,V;if(t[0]!==W||t[1]!==X)...jsxs(...,{color:W,dimColor:X,children:
```

but 2.1.199 emits:

```js
{isLoading:n,isScreenReader:r,themeColor:o}=e,i=o??void 0,a=r?"$":ct.pointer,l;
if(t[0]!==i||t[1]!==n||t[2]!==a)l=yQe.jsxs(w,{color:i,dimColor:n,children:[a,"\xA0"]}),...
```

Three changes broke the regex: (1) a new `isScreenReader:r` field sits between `isLoading` and `themeColor`; (2) an extra `a=r?"$":ct.pointer` assignment appears after `color=themeColor??void 0`; (3) a third `||t[2]!==a` guard term was added.

## Fix
Loosen the anchor to tolerate extra destructured fields (`(?:[$\w]+:[$\w]+,)*`), extra assignments before the `;` (`[^;]*`), and extra guard terms (`[^)]*`) — still capturing isLoading, themeColor, and the resolved-color var. The `color:X,dimColor:Y` → `color:isLoading?resolved:CUSTOM,dimColor:!1` rewrite is unchanged, so behavior is preserved.

## Verification
- Existing tests still pass; added a test for the exact 2.1.199 shape.
- Against the real 2.1.199 native bundle, the rewrite now lands at the chevron:
  `…t[2]!==a)l=yQe.jsxs(w,{color:n?i:"ansi:blue",dimColor:!1,children:[a,…`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility of the input-chevron color patch with newer Claude Code patterns (including screen-reader-related state and additional guard terms), ensuring the correct `color`/`dimColor` rewrite still applies.
* **Tests**
  * Added a Vitest case covering the newer input-chevron pattern to prevent regressions in the patched color logic.
* **Documentation**
  * Updated the Unreleased changelog entry to reflect the input-chevron color fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- Fix input-chevron-color patch for Claude Code 2.1.199 (chevron component gained an `isScreenReader` field and a third guard term) (#855) - @StreamDemon
+
 ## [v4.3.0](https://github.com/Piebald-AI/tweakcc/releases/tag/v4.3.0) - 2026-06-29
 
 - Add input chevron color patch (#634) - @VitalyOstanin
--- a/src/patches/inputChevronColor.test.ts
+++ b/src/patches/inputChevronColor.test.ts
@@ -32,8 +32,6 @@ describe('writeInputChevronColor', () => {
   });
 
   it('handles the CC 2.1.199 shape (isScreenReader field + extra pointer var + 3-term guard)', () => {
-    // 2.1.199 inserts `isScreenReader:r` into the destructure, an extra
-    // `a=r?"$":ct.pointer` assignment, and a third `||t[2]!==a` guard term.
     const input =
       'var z=1,{isLoading:n,isScreenReader:r,themeColor:o}=e,i=o??void 0,a=r?"$":ct.pointer,l;' +
       'if(t[0]!==i||t[1]!==n||t[2]!==a)' +
--- a/src/patches/inputChevronColor.test.ts
+++ b/src/patches/inputChevronColor.test.ts
@@ -31,6 +31,20 @@ describe('writeInputChevronColor', () => {
     expect(result).toBeNull();
   });
 
+  it('handles the CC 2.1.199 shape (isScreenReader field + extra pointer var + 3-term guard)', () => {
+    // 2.1.199 inserts `isScreenReader:r` into the destructure, an extra
+    // `a=r?"$":ct.pointer` assignment, and a third `||t[2]!==a` guard term.
+    const input =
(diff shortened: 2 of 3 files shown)


xRevenantPulse:
Addressed the review nitpicks:

- Removed the explanatory comments in \`inputChevronColor.ts\` and the test (1789883).
- **Keeping the `[^;]*` / `[^)]*` spans.** They are tightly bounded by the surrounding literals (`…??void 0,` … `;if(`, and inside the `if(...)`), and the matcher is verified against the JS extracted from the 2.1.199 native binary — the rewrite lands only at the chevron's `color:…,dimColor:…` site. A stricter pattern would have to hard-code 2.1.199's exact `isScreenReader:r,` field and `a=r?"$":ct.pointer,` assignment, which would break on the next minor bundle shift; the bounded spans deliberately tolerate those while still requiring the specific `isLoading`/`themeColor`/`void 0`/`if([0]!==…||[1]!==…)`/`jsxs(…,{color:…,dimColor:…,children:` anchors. The failure mode is a clean no-match (patch reports ✗), not a mis-injection.

bl-ue (approved):

PR #856, opened 2026-07-03
Fix agents-md async reader anchor for CC 2.1.199 (reader refactored into stat+read helper)

### LINKED ISSUE #851: agents-md patch broken on CC 2.1.199 (async CLAUDE.md reader refactored)
## Environment
- tweakcc: 4.3.0
- Claude Code: 2.1.199 (native binary)
- OS: Linux x64

## Symptom
`agents-md` ("AGENTS.md (and others)") reports ✗ in the apply summary. AGENTS.md / alternative memory filenames are not picked up.

## Root cause (anchor break — async reader refactored)
`writeAgentsMdAsync` in `src/patches/agentsMd.ts` expects the async CLAUDE.md reader to read the file inline:

```js
async function NAME(A,q,K){try{let z=await FS().readFile(A,{encoding:"utf-8"});return PROC(z,A,q,K)}catch(_){return H(_,A),{info:null,includePaths:[]}}}
```

In 2.1.199 that function has been refactored — the read moved into a helper and a regular-file/size guard was added:

```js
async function aya(e,t,n){
  try{
    let r=Vt(),o=await aN(r,e,Ypo);
    if(o===null)return T(`[CLAUDE.md] skipping ${e}: not a regular file or exceeds ${Ypo} byte limit`),{info:null,includePaths:[]};
    ...
  }catch(...){...}
}
```

There is no inline `await …readFile(…,{encoding:"utf-8"})` in this function anymore, so the async pattern misses and returns `null`. The code then falls back to `writeAgentsMdSync`, whose `fsPattern` (`readFileSync|existsSync|statSync`) also finds nothing in the async-only function → final error `patch: agentsMd: failed to find fs expression in function or caller`.

## Suggested fix
Re-anchor on the new `aya`/`aN(r,e,Ypo)` shape (rerouting to the alt-filename lookup from the `if(o===null)` / catch return path). Exact strategy TBD; I'll work it out against the extracted bundle.

## Verification
Ran `writeAgentsMd` against the JS extracted from the 2.1.199 native binary → returns `null`.

Worked on 2.1.185 (✓ in #846's screenshot) — fresh regression in 2.1.186→2.1.199.

Happy to open a PR.

StreamDemon:
Fixes #851.

## Problem
`agents-md` ("AGENTS.md (and others)") reports ✗ on CC 2.1.199 (native binary); alternative memory filenames aren't picked up. (Worked on 2.1.185 — a fresh regression.)

## Root cause
`writeAgentsMdAsync` expected the async CLAUDE.md reader to `readFile` inline. On 2.1.199 the read was refactored into a helper and gained a regular-file/size guard:

```js
async function aN(e,t,n){let r=await e.stat(t);if(!r.isFile()||r.size>n)return null;return await e.readFile(t,{encoding:"utf8"})}

async function aya(e,t,n){try{let r=Vt(),o=await aN(r,e,Ypo);
  if(o===null)return T(`[CLAUDE.md] skipping ${e}: ...`),{info:null,includePaths:[]};
  return FHp(o,e,t,n)}catch(r){return jHp(r,e),{info:null,includePaths:[]}}}
```

There's no inline `readFile(…,{encoding:"utf-8"})` in `aya` anymore, so the old async pattern missed → the code fell back to the sync matcher, which also found nothing (`failed to find fs expression`).

A **missing** CLAUDE.md makes `aN`'s `stat` throw ENOENT, which propagates to `aya`'s `catch` — so that's where the AGENTS.md reroute belongs (the `o===null` branch is only the not-a-regular-file / too-big case).

## Fix
Add `writeAgentsMdAsync2199`, tried before the older async and sync matchers. It matches the new `aya` shape (capturing the function name, path param, pass-through params, catch var, and error handler via backreferences tying the structure together), then injects the reroute loop into the catch and adds a `didReroute` guard param — the same rerouting strategy the patch has always used. Older async/sync installs are untouched (existing matchers remain as fallbacks).

## Verification
- Existing sync-path tests still pass; added a test for the 2.1.199 `aya` shape.
- Against the real 2.1.199 native bundle the patch applies (Δ +233) and produces a well-formed function:

```js
async function aya(e,t,n,didReroute){try{let r=Vt(),o=await aN(r,e,Ypo);if(o===null)return T(`[CLAUDE.md] skipping ${e}: ...`),{info:null,includePaths:[]};return FHp(o,e,t,n)}catch(r){if(!didReroute&&(e.endsWith("/CLAUDE.md")||e.endsWith("\\CLAUDE.md"))){for(let alt of ["AGENTS.md",...]){let altPath=e.slice(0,-9)+alt;try{let _r=await aya(altPath,t,n,true);if(_r.info)return _r}catch{}}}return jHp(r,e),{info:null,includePaths:[]}}}
```

(braces balanced, original branches preserved, recursion guarded by `didReroute`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `CLAUDE.md` patching for newer Claude Code (e.g., CC 2.1.199) with reroute support when reads fail, and stronger handling for alternate filename/path endings.
  * Added a version-specific async patch step before the existing async/sync fallback chain.
* **Tests**
  * Added coverage for the CC 2.1.199 async-reader scenario, asserting signature updates and correct reroute/fallback behavior.
* **Documentation**
  * Updated the changelog under **Unreleased** with the latest fix entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- Fix agents-md patch for Claude Code 2.1.199 (the async CLAUDE.md reader was refactored into a stat+read helper) (#856) - @StreamDemon
+
 ## [v4.3.0](https://github.com/Piebald-AI/tweakcc/releases/tag/v4.3.0) - 2026-06-29
 
 - Add input chevron color patch (#634) - @VitalyOstanin
--- a/src/patches/agentsMd.test.ts
+++ b/src/patches/agentsMd.test.ts
@@ -62,10 +62,6 @@ describe('agentsMd', () => {
       expect(result).toBeNull();
     });
 
-    // CC 2.1.199 refactored the async reader: the read moved into a helper
-    // (aN, which stats then readFiles) and a regular-file/size guard was added.
-    // A missing CLAUDE.md makes aN's stat throw ENOENT → the reader's catch, so
-    // the AGENTS.md reroute must be injected there.
     const asyncReader199 =
       'async function aya(e,t,n){try{let r=Vt(),o=await aN(r,e,Ypo);' +
       'if(o===null)return T(`[CLAUDE.md] skipping ${e}: not a regular file or exceeds ${Ypo} byte limit`),{info:null,includePaths:[]};' +
--- a/src/patches/agentsMd.test.ts
+++ b/src/patches/agentsMd.test.ts
@@ -61,5 +61,33 @@ describe('agentsMd', () => {
       const result = writeAgentsMd('not a valid file', altNames);
       expect(result).toBeNull();
     });
+
+    // CC 2.1.199 refactored the async reader: the read moved into a helper
+    // (aN, which stats then readFiles) and a regular-file/size guard was added.
+    // A missing CLAUDE.md makes aN's stat throw ENOENT → the reader's catch, so
+    // the AGENTS.md reroute must be injected there.
+    const asyncReader199 =
+      'async function aya(e,t,n){try{let r=Vt(),o=await aN(r,e,Ypo);' +
+      'if(o===null)return T(`[CLAUDE.md] skipping ${e}: not a regular file or exceeds ${Ypo} byte limit`),{info:null,includePaths:[]};' +
+      'return FHp(o,e,t,n)}catch(r){return jHp(r,e),{info:null,includePaths:[]}}}';
+
+    it('handles the CC 2.1.199 async reader (reroute injected into catch)', () => {
+      const result = writeAgentsMd(asyncReader199, altNames);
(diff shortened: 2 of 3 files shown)


xRevenantPulse:
Addressed the review nitpicks:

- Removed the added JSDoc/inline comments in \`agentsMd.ts\` and the test comment (a03a021).
- **Keeping the two matchers self-contained** rather than extracting a shared reroute builder. `writeAgentsMdAsync2199` and `writeAgentsMdAsync` target *different, independently-evolving* Claude Code code shapes; the reroute snippets are similar but not identical (temp var `_r` vs `r`, different pass-through params). Given CC has already refactored this reader three times (sync → async → async2199), keeping each version's emitted code local means a future matcher can be added, changed, or deleted in isolation without threading a shared string-builder through all of them. The duplication is ~3 lines and localized, so I judged the coupling cost higher than the DRY benefit here.

bl-ue (approved):

PR #864, opened 2026-07-08
fix: match try/catch agent-color write in session-color patch

VitalyOstanin:
Claude Code 2.1.204 rewrote `saveAgentColor` to wrap the transcript write in `try/catch` and moved the session-id check into a separate `if`, so the `patchSaveAgentColor` pattern (which expected `if(await Z(...),A===W())`) no longer matched and the whole session-color patch returned `null`.

This adds the new try/catch shape as an alternative pattern while keeping the previous if-based pattern for older Claude Code versions. Both shapes expose the same capture groups, so extraction is unchanged.

A test covering the 2.1.204 try/catch structure (with a template-literal + object literal inside the catch body) is added.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved color-save patching to handle an async implementation that wraps the write in a `try/catch`, improving compatibility across app variations.
  * Kept the injected save hook and callback behavior consistent so the expected patched flow is preserved.
* **Tests**
  * Added unit coverage for the `try/catch`-based async save scenario to confirm the patch is applied correctly.

### DIFF
--- a/src/patches/sessionColor.test.ts
+++ b/src/patches/sessionColor.test.ts
@@ -11,6 +11,12 @@ const makeAsyncSaveAgentColor = () =>
   'if(await Hv(K,{type:"agent-color",agentColor:$,sessionId:H}),H===V$())' +
   'WA().currentSessionAgentColor=$;c("tengu_agent_color_set",{})}';
 
+const makeTryCatchSaveAgentColor = () =>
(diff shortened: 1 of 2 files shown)


VitalyOstanin:
Applied in b50ebd6 — factored the shared async-function prefix into a single `prefix` string. Matching behavior is unchanged (all sessionColor tests still pass).

bl-ue (approved):

PR #907, opened 2026-07-22
Fix session-memory total-file-limit env var name (CM_SM -> CC_SM)

### LINKED ISSUE #906: Session memory total-file-limit knob reads CM_SM_TOTAL_FILE_LIMIT instead of CC_SM_TOTAL_FILE_LIMIT
The session-memory patch exposes four token/threshold knobs as `CC_SM_*` env vars, but the total-file-limit knob reads a differently-prefixed name.

`src/patches/sessionMemory.ts`:

```js
const perSectionCode = 'Number(process.env.CC_SM_PER_SECTION_TOKENS??2000)';
const totalFileCode = 'Number(process.env.CM_SM_TOTAL_FILE_LIMIT??12000)';
```

`CM_SM_TOTAL_FILE_LIMIT` is `CM_`, not `CC_`. It is the only `CM_SM_*` in the repo; the other four are `CC_SM_*` (`CC_SM_PER_SECTION_TOKENS`, `CC_SM_MINIMUM_MESSAGE_TOKENS_TO_INIT`, `CC_SM_MINIMUM_TOKENS_BETWEEN_UPDATE`, `CC_SM_TOOL_CALLS_BETWEEN_UPDATES`). So setting `CC_SM_TOTAL_FILE_LIMIT` (the name that matches the family) has no effect, and the total-file-limit knob only responds to a variable no one would think to set.

The name is undocumented, so no user is relying on the current spelling.

StreamDemon:
Fixes #906.

The session-memory patch's total-file-limit knob read `process.env.CM_SM_TOTAL_FILE_LIMIT`, while every sibling knob uses the `CC_SM_*` prefix:

```js
const perSectionCode = 'Number(process.env.CC_SM_PER_SECTION_TOKENS??2000)';
const totalFileCode = 'Number(process.env.CM_SM_TOTAL_FILE_LIMIT??12000)';
```

`CM_SM_TOTAL_FILE_LIMIT` is the only `CM_SM_*` in the repo and is undocumented, so nothing reads it today and no user relies on the current spelling. Rename it to `CC_SM_TOTAL_FILE_LIMIT` so the knob matches the family and responds to the env var a user would actually set.

One-character literal fix, verifiable by inspection. Env-var-name regression coverage for the session-memory patch is introduced by #905's new `src/patches/sessionMemory.test.ts`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected session-memory total-file-limit environment variable handling so the configured limit is applied reliably.
  * Documented the configuration correction in the unreleased changelog.


bl-ue (approved):
Haha, great catch!

Issue #869, opened 2026-07-11
`--apply` produces an unparsable cli.js: backslash-doubling from #664 over-escapes \` inside ${...} interpolations


xRevenantPulse:
`tweakcc --apply` writes a `cli.js` that Claude Code cannot load. Bun exits at startup:

```
SyntaxError: Unexpected identifier '$'. Expected ':' in ternary operator.
```

The vanilla bundle parses clean under `node --check`. The patched one does not.

The break is one line: `src/patches/systemPrompts.ts:176`, added by `a516881` ("Preserve system prompt backslashes", #664).

```js
if (delimiter === '"' || delimiter === "'" || delimiter === '`') {
  replacementContent = replacementContent.replace(/\\/g, '\\\\');
}
```

For the template-literal delimiter, that doubles every backslash in the replacement. Template-literal prompt content is not plain text, though. It carries live `${...}` interpolations, and inside them, real escapes like `\`` (an escaped backtick in a nested template).

`System Prompt: Team memory index pointer instructions` has one. Its pointer text embeds a ternary with a nested template:

```
vanilla:  ${l?`\`${y}\``:y}
patched:  ${l?`\\`${y}\\``:y}
```

The `\`` became `\\``: an escaped backslash, then an unescaped backtick. The backtick closes the nested template one character early. `${y}` now sits where the ternary's `:` belongs. That is the error, verbatim.

This prompt is not one tweakcc modifies. It patches from the shipped defaults. Every `--apply` on an affected version produces the broken bundle, whatever the user customized.

## Reproduction

- Any Claude Code version carrying the prompt above. It ships in every snapshot from 2.1.173 up. I reproduced on 2.1.206 and 2.1.207.
- `tweakcc --apply` (the default config triggers it).
- Launch Claude Code, or `node --check` the patched `cli.js`.

Nothing here is specific to one version. Any bundle whose source has `\`` inside a `${...}` interpolation hits it.

## Two more prompts, quieter

`System Prompt: Combined memory index pointer instructions` and `System Prompt: Memory index pointer instructions` carry the same escaped backtick at depth 0, in the literal text rather than inside an interpolation:

```
vanilla:  \`${MC}\`
patched:  \\\`${MC}\\\`
```

Those do not crash. `escapeDepthZeroBackticks` re-escapes the doubled backslash, so the JS stays valid. But the rendered prompt text gains a stray backslash before each backtick, so the applied prompt no longer matches the source. Wrong output, no exception.

[...]

PR #908, opened 2026-07-22
Verify the patched bundle parses with node --check before writing it

### LINKED ISSUE #871: Parse-gate the patched cli.js before repack so invalid JS fails at apply time (not at Claude Code startup)
`--apply` can write a `cli.js` that does not parse, and report success while doing it.

#869 is the concrete case. The backtick over-escaping produced `${l?`\\`${y}\\``:y}`, tweakcc printed *"Your system prompt customizations were still applied successfully"*, wrote the binary, and Claude Code aborted at startup with a `SyntaxError`. Whatever validation exists today did not gate that.

`--apply` does print a per-patch success/failure list, but it reports whether each patch found its anchor and performed its replacement, not whether the assembled `cli.js` parses. The prompt that crashed here was in the succeeded list. So the per-patch report and a parse check are orthogonal: the former can be all green while the output is invalid.

The reason is structural. Between building the patched content and repacking it, there is no parse step. `applySystemPrompts` (and the other patches) hand the final string to `writeContent`, which calls `repackNativeInstallation` (`src/patches/index.ts:980-985`, `src/lib/content.ts:55`, `src/nativeInstallation.ts`). Nothing in that path runs a parser. So a syntactically-broken bundle ships, and the failure surfaces at Claude Code startup instead of at `--apply`, where tweakcc still holds the broken string.

## Proposal

Add a parse check on the fully-patched `cli.js`, right before the repack (around `src/patches/index.ts:981`). If it does not parse, do not write it. Report the parse error instead of shipping the binary.

The check should be `node --check` on the extracted JS (or a real parser in the mode the runtime uses). That is the tool that located #869, at `:522`. The naive in-process options are the wrong instrument here: `new Function(code)` and `vm.compileFunction(code)` both wrap the source in a function body, which false-positives on top-level `import`/`export` or top-level `await`. A gate that misfires is worse than none.

Cost is sub-second on a ~19 MB bundle, negligible next to the repack.

## Open design question: abort vs skip

The gate runs after all patches, so on failure it knows the bundle is broken but not which injection broke it. Two directions:

- Abort the whole `--apply` and restore. Safe, but one stray custom prompt blocks every customization.
- Attribute the failure to a specific patch (bisect, or validate incrementally) and skip only that one.

`applySystemPrompts` already skips individual prompts on the `incomplete` path, but that is a per-prompt escape check, not a parse of the assembled output. A post-apply gate is a different layer. Worth deciding which behavior fits before building it.

## Caveat

Claude Code runs on Bun (JavaScriptCore); tweakcc runs on Node (V8). `node --check` catches the overwhelming majority of invalid JS, including #869 (a ternary/template error both parsers reject), but it is a strong proxy, not a perfect Bun oracle.

## Why it is worth it

This is not specific to #869. The gate catches any patch, code or prompt, that emits invalid JS, and it catches the user-modified case #870's review surfaced: a custom backtick prompt with a stray literal backslash (`C:\path`, a trailing `\`, a `\u`/`\x` not followed by hex) still produces invalid JS that #870 cannot address. A parse check turns all of these from a silent Claude Code crash into a clear apply-time failure.

Motivating case: #869 (fixed by #870).

StreamDemon:
`src/patches/index.ts`: after the patch loop assembles the final `cli.js` and before the repack, run `node --check` on it. If it does not parse, abort the apply without writing and print the syntax error.

Today `--apply` has no parse step between building the patched bundle and repacking it. A patch or a prompt that emits invalid JS ships, `--apply` reports success, and Claude Code aborts at startup with a `SyntaxError`. #869 was the concrete case: the per-patch list was all green and the binary still would not boot.

## Change

- New `src/patches/parseGate.ts`: `assertPatchedBundleParses(content)` writes the content to a temp `.cjs` file and runs `execFileSync(process.execPath, ['--check', tmp])`. A non-zero exit throws `PatchedBundleParseError` carrying a sanitized message: the `SyntaxError` summary and a caret-centered source excerpt, with the temp path, V8 stack, and `node:internal` frames stripped, capped so a corrupted long minified line cannot dump the whole line.
- Wired into `applyCustomization` right before the native repack / npm write. On failure it sets `changesApplied=false` (the binary was already restored to pristine at the start of apply, so nothing shipped) and rethrows.
- `handleApplyMode` reports the failure and exits non-zero. The customizations are not applied; Claude Code runs the original unmodified version.

The extracted bundle is CommonJS (`@bun-cjs`); the `.cjs` extension pins CommonJS parsing regardless of any ambient `package.json` `"type"`. `new Function` and `vm.compileFunction` impose a bare function-body context that diverges from module parsing, so they are not used. `node --check` writes its diagnostic to stderr then exits, which truncates a piped stderr on long lines, so stderr is captured to a file. If the temp file cannot be created or written, the check is skipped rather than blocking an otherwise-valid apply.

## Scope

`node --check` runs on Node (V8); Claude Code runs on Bun (JSC). It is a strong proxy, not a perfect Bun oracle. It catches #869 and the user-modified backtick cases that #870 cannot fix (a stray `C:\path`, a trailing `\`, a `\u`/`\x` not followed by hex).

## Tests

- `parseGate.test.ts`: valid CommonJS passes; `let a=;` and the exact #869 over-escaped template/ternary throw; a break on a 300 KB single line still surfaces a bounded `SyntaxError`; `export`/`import`/`await` inside string literals do not false-positive; `sanitizeParseError` is unit-tested against a contract (non-empty, never leaks the temp path, drops stack/internal/footer noise).
- `modelCustomizationsToggle.test.ts`: the gate runs before the write, and a parse failure aborts the write and marks `changesApplied=false`.

Verified on a real Claude Code 2.1.217 native install and offline on the extracted 20 MB bundle.

Closes #871.

## Summary by CodeRabbit

- **Bug Fixes**
  - `--apply` now validates the fully patched bundle with a Node syntax check before writing changes.
  - If parsing fails, the tool aborts without modifying the installation, prints a concise sanitized syntax error, and exits with code `1`.
- **Tests**
  - Added parse-gate coverage, including regression fixtures and verification that write/replace is skipped and `changesApplied` stays `false` on parse failures.
  - Includes error sanitization and additional edge-case behavior tests.
- **Documentation**
  - Updated the Unreleased changelog entry to reflect the new `--apply` validate/abort behavior.


xRevenantPulse:
Round 2 addressed in `e71651d`.

**index.ts (narrow the catch):** done. The catch now rethrows anything that is not a `PatchedBundleParseError` unchanged, so only a genuine parse failure sets `changesApplied=false`.

**parseGate.ts (openSync fail-open):** done. `openSync` for the stderr file now follows the same skip-with-warning path as `mkdtemp`/`writeFileSync`, matching the documented "check is skipped if the temp file can't be created" contract.

**parseGate.ts (bounded stderr read):** not changing this one. `readFileSync` is bounded by `node --check`'s output for the single erroring line, which is O(longest line in the bundle). The 2.1.217 bundle's longest line is 538 KB, and the full ~20 MB `content` is already resident, so the read adds under 1 MB transiently and is freed after sanitizing. A head-bounded read would stop inside that 538 KB line and drop the trailing `SyntaxError:` summary, reintroducing the pipe-truncation this file-capture path exists to fix (the earlier revision lost the summary at ~146 KB). `sanitizeParseError` already caps the returned message at 2000 chars.

xRevenantPulse:
Round 3 addressed in `37dd89c`.

**parseGate.ts (distinguish parse failures from operational failures):** done. The catch now treats only a genuine non-zero exit as a parse failure (`node --check` sets `status` to a number on a syntax error). A timeout (`status` null, `SIGTERM`), a signal kill, or a spawn failure warns and skips the check, matching the fail-open path already used for temp-file setup. Confirmed the shapes: parse failure is `status=1, signal=null`; timeout is `status=null, signal=SIGTERM, code=ETIMEDOUT`; spawn failure is `status=null, code=ENOENT`.

**modelCustomizationsToggle.test.ts (`@/` import alias):** not changing this one. `src/patches/` uses relative imports almost universally (175 relative specifiers against 1 `@/`). Every import in both the test and `patches/index.ts`, where `applyCustomization` lives, is relative, so `./parseGate` matches the surrounding code. Switching only this one to `@/patches/parseGate` would make it the outlier, and since the test import, the `vi.mock` target, and the production import are all already `./parseGate`, they resolve to a single module identity.

bl-ue (approved):

Issue #899, opened 2026-07-21
System-prompt sync marks a prompt up-to-date on version stamp alone, keeping a stale .md whose identifiers drifted


xRevenantPulse:
## Summary

`syncPrompt` decides a `.md` is up to date by comparing version stamps only (`src/systemPromptSync.ts:869`):

```ts
const versionComparison = compareVersions(existingFile.ccVersion, prompt.version);
if (versionComparison === 0) {
  result.action = 'skipped';
  return result;
}
```

When the extractor renames a prompt's interpolation identifiers in a newer snapshot without bumping that prompt's per-prompt `version`, the stamps stay equal, sync skips, and the stale `.md` persists.

`updateVariables` (`src/systemPromptSync.ts:344`, called unconditionally at `:860`) refreshes the frontmatter `variables:` list to the new names but leaves `parsed.content` (the body) untouched. So the frontmatter advertises the new identifier names while the body still uses the old `${OLD_NAME}` tokens: a silent mismatch that no message surfaces.

## Consequence

On the next `--apply`, the stale body injects an identifier that the current bundle does not define, and Claude Code crashes at runtime (`X is not defined`). See #900 for the crash and the apply-side guard.

## Repro

- A prompt whose `.md` was generated from an older CC version (e.g. `system-prompt-memory-instructions`, `ccVersion: 2.1.205`).
- A newer snapshot (2.1.216) that renamed the prompt's identifiers (`HAS_PROJECT_SKILL_UPKEEP_INSTRUCTIONS` -> `HAS_PROJECT_SKILL_UPKEEP_INSTRUCTIONS_FN`) while keeping the same per-prompt `version` stamp.
- Sync reports "skipped up-to-date"; the body is never regenerated.

## Proposed fix

When the version stamps are equal, also verify the `.md` body is structurally compatible with the current `identifierMap` (every interpolation the current pieces expect resolves to a current identifier). If the body has drifted, regenerate it (or flag a conflict) instead of skipping. This heals the `.md` so the customization survives, rather than relying only on the apply-side guard to skip it.

## Environment

- tweakcc 4.3.2, Claude Code 2.1.216 (native install), Node 24.

PR #911, opened 2026-07-23
Fix system-prompt patching only replacing the first occurrence (#678)

### LINKED ISSUE #678: System prompt patching only replaces first occurrence — misses dual-path prompt arrays in v2.1.104+
## Description

Claude Code v2.1.104 has two code paths for the "Doing tasks" system prompt section — a full-mode array (normal sessions) and a compact-mode array (post-compaction). Several prompt strings appear in BOTH arrays. tweakcc's system prompt replacement only matches and replaces ONE occurrence, leaving the other path with vanilla text.

## Reproduction

1. Customize `system-prompt-doing-tasks-no-unnecessary-error-handling.md` with any change
2. Run `tweakcc --apply --patches "system-prompt-doing-tasks-no-unnecessary-error-handling"`
3. Extract JS: `tweakcc unpack /tmp/check.js`
4. Count occurrences of the vanilla string:

```bash
grep -o "Don't add error handling, fallbacks" /tmp/check.js | wc -l
# Expected: 0
# Actual:   1  (one occurrence remains in full-mode path)
```

The vanilla string exists twice in the unpatched binary (one per code path). tweakcc only replaced the first match.

## Root Cause

The replacement regex uses surrounding context to locate the target string. When multiple system prompt edits are applied, earlier edits change the context around later edit locations. The regex for the second prompt edit no longer matches because the preceding string was already modified by the first edit.

Specifically, in v2.1.104's JS bundle:

```javascript
let _ = isCompactMode(H);
let K = [
  ..._? [
    COMPACT_STRING_1,    // Edits 1+3 condensed
    COMPACT_STRING_2,    // Edit 2
  ] : [
    FULL_STRING_1,       // Edit 1 (changed by tweakcc)
    FULL_STRING_2,       // Edit 2 (context changed, regex misses)
    FULL_STRING_3,       // Edit 3 (changed by tweakcc)
  ]
];
```

After Edit 1 is applied, the context before Edit 2's location in the full-mode path changes from the vanilla text to the customized text. The regex no longer matches.

## Affected Prompts (v2.1.104)

The dual-path structure appears for these "Doing tasks" subsections:
- `system-prompt-doing-tasks-no-unnecessary-error-handling` — vanilla exists in both paths
- `system-prompt-doing-tasks-no-additions` + `system-prompt-doing-tasks-no-premature-abstractions` — compact path has a CONDENSED version (different text from individual files)
- `system-prompt-communication-style` — compact path has a different expanded version

## Workaround

After `tweakcc --apply`, extract JS, do a global string replacement with Python (`str.replace()` matches all occurrences regardless of context), and repack:

```bash
tweakcc unpack /tmp/fix.js
python3 -c "
with open('/tmp/fix.js', 'r') as f: content = f.read()
content = content.replace(OLD_VANILLA_TEXT, NEW_CUSTOM_TEXT)
with open('/tmp/fix.js', 'w') as f: f.write(content)
"
tweakcc repack /tmp/fix.js
```

## Suggested Fix

Consider replacing all occurrences of the vanilla string rather than only the first regex match. The system prompt text is unique enough that a global `str.replace()` is safe — the vanilla string won't appear elsewhere in the bundle.

Alternatively, document the dual-path architecture so users know to handle both paths.

## Additional Note: --patches flag

The `--patches` flag is an excellent workaround for issue #676 (code patches failing on v2.1.104). By whitelisting only system prompt patch IDs, all "Always Applied" code patches are skipped, resulting in zero failures. Users can apply system prompt customizations cleanly even when code patches are broken. This workaround might be worth documenting in the README.

## Environment

- Claude Code: v2.1.104 (native binary, macOS x86_64)
- tweakcc: v4.0.11
- Node: v22.x

arimu1:
## Bug

`applySystemPrompts` compiled each prompt's search regex without the `g` flag:

```js
pattern = new RegExp(regex, 'si');
match = content.match(pattern);
...
content = content.replace(pattern, () => replacementContent);
```

`cli.js` sometimes repeats the exact same prompt text at more than one location — e.g. Claude Code's full-mode vs. compact-mode "Doing tasks" prompt arrays. Since `.replace()` with a non-global regex only replaces the first match, the first occurrence got patched and any later occurrence(s) were silently left vanilla.

Reproduction from #678: customize `system-prompt-doing-tasks-no-unnecessary-error-handling`, apply, unpack the resulting bundle, and the vanilla string is still findable once even though it existed twice pre-patch.

## Fix

- The search regex now compiles with the `g` flag, and all matches are collected via `content.matchAll(pattern)` up front.
- Each occurrence is interpolated and escaped **independently** rather than computing the replacement once and reusing it for every occurrence. This matters because the minified variable names captured at each call site (used for `${...}` interpolation) can differ between occurrences even though the surrounding prompt text is byte-identical — a plain global string replace would silently paste the first occurrence's variable name into the second occurrence, corrupting it. The existing per-match validation for stale/renamed interpolation identifiers (#899/#900) and incomplete-backtick-escaping guard are likewise now run per-occurrence, and the whole prompt is left unpatched (as before) if any occurrence fails those checks.
- `details` now notes the occurrence count (e.g. `"12 fewer chars (2 occurrences)"`) when a prompt was patched in more than one place.

## Tests

Added `describe('patching all occurrences of a repeated prompt (#678)', ...)` to `src/patches/systemPrompts.test.ts`:
1. A prompt with vanilla text repeated in two code paths — asserts both are patched and zero vanilla occurrences remain.
2. A prompt with variable interpolation where the two occurrences have *different* captured variable names — asserts each occurrence keeps its own variable rather than the first match's being reused for both.

Both tests fail against the pre-fix code (verified by stashing the `systemPrompts.ts` change and re-running): the first shows the second occurrence still vanilla, the second shows the second occurrence entirely unpatched.

```
pnpm run test
 Test Files  43 passed (43)
      Tests  458 passed | 5 skipped (463)

pnpm lint
$ tsc --noEmit && eslint src
(no errors)
```

Fixes #678.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed system-prompt patching to update all matching occurrences, not only the first.
  * Correctly applies per-occurrence interpolations with proper captured values.
  * Improved escaping/validation across supported string formats and prevents unintended prompt changes.
  * Updated patch status messaging to include how many occurrences were processed.
* **Tests**
  * Added coverage for multi-occurrence patching and per-occurrence interpolation behavior.
* **Documentation**
  * Updated the changelog with the fix under “Unreleased.”


bl-ue:
Hi @arimu1, I merged a few more PRs before yours and now there are a couple conflicts—will you please fix them so I can merge?

arimu1:
Rebased onto latest `main` and resolved conflicts in `CHANGELOG.md` and `src/patches/systemPrompts.ts` (kept multi-occurrence replace + non-ASCII encoding order from #920). `vitest` systemPrompts suite: 48/48 pass.

arimu1:
@bl-ue Rebased/synced onto current `main` — GitHub now reports the PR as **mergeable** with no conflicts. Ready for another look when you have a chance.

arimu1:
@bl-ue thanks — branch is rebased/synced on current `main` and GitHub reports the PR as mergeable again. Ready for another look whenever you are.

bl-ue (approved):

PR #913, opened 2026-07-23
Restore the session-memory update-cadence env var on Claude Code 2.1.218 and mark the token-limit knobs legacy-only

### LINKED ISSUE #912: Session-memory CC_SM_* tuning knobs silently no-op on Claude Code 2.1.218
## Summary

Four of the five `CC_SM_*` session-memory env vars silently do nothing on Claude Code 2.1.218. `--apply` reports Session memory as applied (✓), but setting `CC_SM_PER_SECTION_TOKENS`, `CC_SM_TOTAL_FILE_LIMIT`, `CC_SM_MINIMUM_MESSAGE_TOKENS_TO_INIT`, or `CC_SM_MINIMUM_TOKENS_BETWEEN_UPDATE` has no effect, with no warning.

## Why

Claude Code's ~2.1.217 memory-model refactor replaced the single session-memory-file model with the multi-file memory system. The constructs those four knobs patch are absent from the 2.1.218 bundle:

- `patchTokenLimits` anchors on `=2000 ... =12000 ... # Session Title`. The `# Session Title` landmark has 0 occurrences in 2.1.218 (the file-format prompt now uses `# Memory` / `## Types of memory` headings).
- `patchUpdateThresholds` anchors on `minimumMessageTokensToInit:1e4`, `minimumTokensBetweenUpdate:5000`, `toolCallsBetweenUpdates:3`. All three identifiers have 0 occurrences in 2.1.218.

On the refactored path `usedLegacyExtraction` is false, so both sub-patches return null non-fatally (no crash, no log). The feature still applies via the extraction/past-session patches, so Session memory shows ✓ while these knobs quietly do nothing. This is the same "reports success while not applying" shape as #888/#889.

The one update-cadence knob that survives is `toolCallsBetweenUpdates`, whose role is now served by the GrowthBook flag `getFlag("tengu_bramble_lintel", null) ?? 1`.

## Expected

- `CC_SM_TOOL_CALLS_BETWEEN_UPDATES` should configure the surviving cadence on 2.1.218.
- The four token-limit/threshold knobs should be documented as legacy-only (they apply only to pre-refactor Claude Code) rather than appearing to work.

## Environment

Claude Code 2.1.218 (native install), tweakcc built from `main`. Verified by extracting the bundle and grepping: `# Session Title` = 0, `minimumMessageTokensToInit` = 0, `tengu_bramble_lintel` = 1 (`...??1`).

StreamDemon:
Fixes #912.

## What

Claude Code's ~2.1.217 memory-model refactor removed the declarative update-threshold object and the `# Session Title` per-section/total token budgets, so four of the five `CC_SM_*` session-memory env vars no longer have an anchor on current bundles and silently no-op (the feature still applies, so Session memory reports ✓ while the knobs do nothing).

This restores the one knob that survives and makes the rest honest:

- **Re-anchor the survivor.** The update cadence is now gated by `getFlag("tengu_bramble_lintel", null) ?? <n>`. `patchUpdateThresholds` gains a match that rewrites the numeric default to `Number(process.env.CC_SM_TOOL_CALLS_BETWEEN_UPDATES ?? <n>)`, preserving the flag's precedence and carrying the upstream default through (via `$2`) instead of hard-coding it. So `patchUpdateThresholds` now reports honest success on 2.1.218 instead of silently returning null.
- **Honest-gate the rest.** `patchTokenLimits` and the three legacy `minimum*/toolCalls` replacements are documented as legacy-only (pre-refactor Claude Code). They already no-op non-fatally there; the comments make that a deliberate, documented limitation rather than a silent gap.

## Notes / caveats

- **Env-var name.** `CC_SM_TOOL_CALLS_BETWEEN_UPDATES` is retained for continuity with the legacy knob. The `tengu_bramble_lintel` flag gates an extraction-cycle cadence (`if(!u){if(o++,o<g)return}`), not a literal tool-call count, so the name is now approximate. Open to a rename if preferred.
- **This is not the 2.1.217 fatal fix.** #905 defused the fatal break by fixing the legacy-extraction classification. This PR is separate: it restores a tuning knob on the refactored bundle and stops the silent no-op.
- **Honest-gating is documentation, not a runtime warning.** These env vars are read inside the injected code at Claude Code runtime, not at patch time, so tweakcc cannot warn a user who sets a legacy-only var (the injection site is gone). Documenting them as legacy-only is the available honest signal.

## Verification

- 9 unit tests in `sessionMemory.test.ts` (RED before, GREEN after): re-anchor on the 2.1.218 shape, default carry-through (`??4`), a negative test that other `tengu_*` cadence flags are untouched, a negative test that a non-integer default is rejected, plus the existing legacy/telemetry cases.
- Real-bundle check: `writeSessionMemory` on the pristine extracted 2.1.218 `cli.js` re-anchors the single `tengu_bramble_lintel` site; output passes `node --check`.
- Live end-to-end on a real 2.1.218 native install: `--apply -y` shows Session memory ✓ with no failures, the patched binary carries `...??Number(process.env.CC_SM_TOOL_CALLS_BETWEEN_UPDATES??1)`, `node --check` passes, it boots, and `-p` runs a turn cleanly.
- Full suite (460 pass / 5 skip), lint clean.

## Summary by CodeRabbit

* **New Features**
  * Restored session-memory update cadence control for Claude Code 2.1.218, with correct re-anchoring to the current cadence mechanism.

* **Documentation**
  * Clarified that legacy session-memory token-limit and update-threshold knobs are legacy-only, while cadence-based control remains supported.

* **Bug Fixes**
  * Fixed session-memory environment-variable rewriting to preserve cadence/flag precedence and numeric defaults, avoid rewriting when cadence defaults aren’t whole numbers, and prevent unintended changes to unrelated cadence flags.


xRevenantPulse:
@coderabbitai Nit is a valid finding, applied with one adaptation.

The proposed positive lookahead `(?=[,;}])` doesn't survive our `globalReplace` helper: it re-applies the pattern to the isolated matched substring (`matchedStr.replace(singlePattern, ...)`), and `match[0]` excludes the zero-width lookahead, so the trailing `[,;}]` is never present in the slice and the replacement silently no-ops. The existing legacy regression test caught it (the whole patch returned null). `\b` and negative lookaheads survive that re-application because they are satisfied at end-of-string; a positive lookahead is not.

Consuming the trailing delimiter and restoring it via a second capture group gives the same literal-delimiter compliance and works with `globalReplace`:

```
/([,{;])minimumMessageTokensToInit:1e4([,;}])/g,
'$1minimumMessageTokensToInit:Number(process.env.CC_SM_MINIMUM_MESSAGE_TOKENS_TO_INIT??1e4)$2'
```

Applied to all three legacy regexes, plus an assertion that the object delimiters round-trip (`...??1e4),minimumTokensBetweenUpdate:...??3)}`).

xRevenantPulse:
Re-verified on Claude Code 2.1.220, off current `main` (`48ac856`) with this branch applied. The patch applies unchanged.

The injected site in the patched bundle:

```js
Ke("tengu_bramble_lintel",null)??Number(process.env.CC_SM_TOOL_CALLS_BETWEEN_UPDATES??1)
```

Note the callee. It minified to `Xe` on 2.1.218 and to `Ke` on 2.1.220. The match keys off the call shape rather than the identifier, so the rename cost nothing.

`tengu_bramble_lintel` occurs exactly once in the pristine 2.1.220 bundle and exactly once in the patched one, so there is no second cadence path sitting unpatched behind the first.

The legacy-only claims hold on 2.1.220 as written. `# Session Title` and `minimumMessageTokensToInit` are both absent from the pristine binary, and `CC_SM_TOTAL_FILE_LIMIT` does not appear in the patched JS.

Live apply on a real 2.1.220 native install: 0 failures across the report, Session memory ✓, all 15 code patches applied, `node --check` passes, it boots, `-p` runs a turn.

Rebased onto current `main` (`48ac856`). Single commit, unchanged content, still touching only `CHANGELOG.md`, `src/patches/sessionMemory.ts` and its test.

bl-ue (approved):

PR #915, opened 2026-07-23
Place the repacked .bun right after the writable ELF segment to remove the ~262MB zero-padding gap

### LINKED ISSUE #914: Native --apply bloats the patched ELF binary with a ~262MB zero-padding gap (273MB to 721MB on 2.1.218)
## Summary

On a native (Bun-compiled ELF) install, `tweakcc --apply` produces a patched binary far larger than needed. On Claude Code 2.1.218 the pristine 273MB binary becomes **721MB (~2.8x), 58.6% of it zero bytes**. It is a size inefficiency, not a correctness bug (it boots and runs), but it wastes ~448MB of disk per patched install.

## Root cause

`repackELFSection` places the rebuilt `.bun` section at `LIEF.nextVirtualAddress()`, then extends the writable `PT_LOAD` to reach it. Two things inflate the file:

1. **~262MB zero-padding gap.** `nextVirtualAddress()` rounds up to a coarse boundary: on 2.1.218 the highest LOAD segment ends at vaddr `0x1067a000` (~275MB) but `nextVirtualAddress()` returns `0x20000000` (512MB). A `PT_LOAD` maps a contiguous file->vaddr range, so extending the segment to reach it materializes the vaddr gap as real zero bytes (measured: longest zero run = 261,645,486 bytes; the `.bun` lands at file offset `0x1fe00000`).
2. **~187MB orphaned old `.bun`.** The original `.bun` (including the unpatched cli.js) stays at its old offset, never reclaimed. The one patched binary contains both the unpatched and patched forms of the same code.

ELF-only: `repackMachO` extends the `__BUN` segment in place by the size delta and `repackPE` sets the section size in place; neither balloons.

## Verified on

Claude Code 2.1.218 native ELF (WSL/Linux), via `readelf -lW` / `-SW` plus a byte-level zero-run scan; a `node-lief` probe confirms `nextVirtualAddress()` = `0x20000000` vs segment end `0x1067a000`.

## Fix direction (PR to follow)

Placing the rebuilt `.bun` immediately after the writable segment (page-aligned) when that segment is topmost cuts the 2.1.218 patched binary from 721MB to ~460MB with byte-identical embedded JS. The remaining ~187MB (orphaned old `.bun`) is a separate, harder step: reclaiming it means rewriting inside the segment's original file range. Worth a follow-up, or would you rather rethink the repack layout wholesale? Happy to take direction.

StreamDemon:
Fixes #914 (partially; see scope).

## What

On native (Bun-compiled ELF) installs, `repackELFSection` placed the rebuilt `.bun` at `LIEF.nextVirtualAddress()` (which rounds up to a coarse ~256MB boundary) and extended the writable `PT_LOAD` to reach it. A `PT_LOAD` maps a contiguous file range, so the vaddr gap became real zero bytes: the 2.1.218 patched binary ballooned from 273MB to 721MB, 58.6% zeros.

This places the new section immediately after the writable segment's mapped end (page-aligned) when that segment is the topmost LOAD segment, so extending it cannot overlap a higher segment. It falls back to the previous `nextVirtualAddress()` placement otherwise (byte-identical behavior for non-topmost layouts).

## Scope: reads 721MB -> 460MB, not -> 273MB

This removes **only** the ~262MB zero-padding gap. The other ~187MB is the original `.bun` module data left orphaned in the file (the old code never reclaimed it either); reclaiming that means rewriting inside the segment's original file range and is deferred to a follow-up. **~460MB is the intended result of this change**, not a sign it is half-broken.

ELF-only: `repackMachO` extends in place by the size delta and `repackPE` sets the section size in place; neither balloons, and `nextVirtualAddress()` is used nowhere else.

## How

Extracted the placement math into a pure, exported `computeBunSectionPlacement` helper (the risky arithmetic, away from LIEF) and unit-tested its invariants: skew preserved (`newVaddr - newFileOffset == rwVaddr - rwFileOffset`, the ELF mapping requirement), no overlap (`newVaddr >= topmost LOAD end`), gap reclaimed, page alignment (compact and fallback), the topmost-segment guard, and memsz-not-filesz for the segment end. `repackELFSection`'s extend / section-field / `patchAddress` / write path is unchanged.

## Verification

- 8 unit tests for the helper (RED -> GREEN), including mutation-guarded cases: a filesz-instead-of-memsz mutation flips the guard and is caught; unaligned mem-end / unaligned `nextVirtualAddress` cases pin the page alignment.
- Real-binary E2E on a copy of the pristine 2.1.218 ELF: `--apply` yields a 460MB binary (down from 721MB), the `.bun` lands at vaddr `0x1067a000` (no gap), the JS extracted from the 460MB binary is **byte-identical** to what tweakcc embedded, `node --check` passes, it boots, and `-p` runs a turn. The apply report is clean (0 failures, all patches applied).
- Full suite (462 pass / 5 skip), lint clean.
- Caveat: the real 2.1.218 RW segment has memsz == filesz, so the memsz != filesz (BSS-gap) path is exercised by unit tests but not on a real binary.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved native Linux/ELF `--apply` by inserting the rebuilt `.bun` section right after the writable segment instead of using a rounded placement, greatly reducing rebuilt binary size and avoiding large zero-padding gaps.
  * Enhanced placement to prevent overlaps with existing ELF segments and to keep correct virtual/file offsets with proper alignment/page rounding.
  * Correctly accounts for memory sizing to avoid issues near writable/BSS tails.
* **Tests**
  * Added Vitest coverage for compact vs fallback `.bun` placement, non-overlap, and alignment/page-rounding behavior.
* **Documentation**
  * Added an Unreleased changelog entry describing the improved native `--apply` behavior.


xRevenantPulse:
`node-lief` moved to v1.3.2 in #916, after this PR was opened. That bump is in exactly the area this change works around, so I re-probed the placement math against it before assuming the workaround still earns its place.

It still over-rounds.

Pristine Claude Code 2.1.220, node-lief 1.3.2:

```
RW PT_LOAD (topmost) end = 0x1083a000    264.2 MB
nextVirtualAddress()     = 0x20000000    512.0 MB
gap                                      247.8 MB
```

`nextVirtualAddress`, `segments`, `extend` and `getSection` keep their 1.3.0 signatures and semantics under 1.3.2, so `computeBunSectionPlacement` applies unchanged and is not double-correcting a value LIEF now handles itself.

Re-verified end to end on a real 2.1.220 native install, off current `main` (`48ac856`) with this branch applied. `--apply -y` reports 0 failures with all 15 code patches applied. The patched binary is 463,698,928 B. The new `.bun` lands at vaddr `0x1083a000`, which is exactly the pristine writable-segment end, so the compact path is the one being taken. `node --check` passes, it boots, `-p` runs a turn.

One number stated precisely: the ~692MB counterfactual for 2.1.220 (512MB placement plus the 180MB section) is arithmetic, not a measurement. The measured before/after pair is still the 2.1.218 one in the PR body, 721MB to 460MB.

Worth noting for anyone sizing the remaining gap: pristine 2.1.220 is already 32.2% zeros, and the patched binary is 35.1%. Zero ratio on its own was never the signal here. The `nextVirtualAddress() - rwEnd` delta is.

Rebased onto current `main` (`48ac856`). Single commit, unchanged content, still touching only `CHANGELOG.md`, `src/nativeInstallation.ts` and its test.

mike1858 (approved):

signadou (approved):

bl-ue (approved):

PR #918, opened 2026-07-24
fix: match refactored jsx chevron component on CC 2.1.219

VitalyOstanin:
Claude Code 2.1.219 refactored the input chevron component: the render switched from `jsxs(...)` to `jsx(...)`, the chevron symbol was hoisted into its own memoized block (with its own braces) between the resolved-color init and the render, and the render's memo guard moved to indices `[2]`/`[3]`/`[4]`. The old pattern skipped up to the guard with `[^{}]*?`, was anchored to guard indices `[0]`/`[1]`, and required `jsxs`, so it no longer matched.

This relaxes the pattern in three ways:

- skip the intermediate memo block with `[\s\S]*?` instead of `[^{}]*?`
- drop the `[0]`/`[1]` index anchors from the memo guard
- accept either `jsx` or `jsxs`

Verified against extracted bundles for 2.1.195, 2.1.201, 2.1.204, 2.1.214 and 2.1.219: on every version where the old pattern matched, the new pattern produces a byte-identical result, and 2.1.219 now matches where the old pattern failed. The relaxed skip does not cost performance — in the worst case (already-patched 20MB bundle, no tail to match) the new pattern runs in ~14ms versus ~28ms for the old one.

A test covering the 2.1.219 shape is added alongside the existing 2.1.199 and 2.1.204 tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility when patching input components built with the CC 2.1.219 compiled format.
  * Refined chevron and theme color rewriting to handle additional render patterns reliably.

* **Tests**
  * Added coverage for the updated chevron/theme color transformation, including ensuring the correct conditional color form is produced and unpatched patterns are not left behind.

### DIFF
--- a/src/patches/inputChevronColor.test.ts
+++ b/src/patches/inputChevronColor.test.ts
@@ -56,6 +56,20 @@ describe('writeInputChevronColor', () => {
     expect(result).not.toContain('color:ZFp,dimColor:QFp');
   });
 
+  it('handles the CC 2.1.219 shape (jsx render + intermediate memo block with braces + guard on [2]/[3])', () => {
+    const input =
+      'var z=1,{isLoading:YVf,isScreenReader:XVf,themeColor:VmI}=qmI,JVf=VmI??void 0,Ytl;' +
+      'if(z1S[0]!==XVf)Ytl=XVf?U2.jsx(U2.Fragment,{children:"$\\xA0"}):' +
+      'U2.jsxs(U2.Fragment,{children:[je.pointer,"\\xA0"]}),z1S[0]=XVf,z1S[1]=Ytl;else Ytl=z1S[1];let K1S;' +
+      'if(z1S[2]!==JVf||z1S[3]!==YVf||z1S[4]!==Ytl)' +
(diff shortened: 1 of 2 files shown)


VitalyOstanin:
Thanks — the finding is valid and is fixed in f126ecb.

The old `[^{}]*?` skip could not contain `{`, and the `color:<resolved>,dimColor:<isLoading>` pair only ever appears inside a jsx props object, so the previous pattern was structurally protected against a duplicate pair. Relaxing the skip to `[\s\S]*?` removed that protection while the replacement still used `indexOf`, so a repeated pair inside the skipped block would have been rewritten instead of the chevron render.

Since the match always ends at the intended pair (`...{color:\3,dimColor:\1,children:`), `lastIndexOf` targets it unambiguously. Confirmed on a fixture with a duplicate pair: with `indexOf` the intermediate block is rewritten and the real render is left untouched; with `lastIndexOf` only the final pair changes.

A regression test with a duplicate pair inside the skipped block is added, asserting the earlier occurrence stays intact.

No behaviour change on real bundles: for 2.1.195, 2.1.201, 2.1.204, 2.1.214 and 2.1.219 the first and last occurrence indices are equal and the patched output is byte-identical either way.

bl-ue (approved):

Issue #921, opened 2026-07-25
System-prompt sync writes undecoded `\"` into the generated `.md`, so the markdown shows JS escape syntax and `--apply` embeds a spurious backslash


xRevenantPulse:
`~/.tweakcc/system-prompts/data-claude-api-reference-csharp.md` contains 66 occurrences of the two-character text `\"` and zero bare `"` characters:

```
…`Betas = [\"code-execution-2025-08-25\", \"skills-2025-10-02\"]`…
```

Those backslashes are JavaScript string-literal syntax from the bundle. They are not part of the prompt. Sync copies them into the markdown instead of decoding them, so anyone opening that file to customise it sees escape syntax rather than the text Claude Code actually receives.

28 of the 610 files generated for Claude Code 2.1.220 are affected.

The same decode gap shows up in a smaller variant: `skill-artifact-whiteboard.md` and `skill-artifact-data-table.md` carry literal `\uXXXX` text for the same reason. A fix should cover both spellings rather than quotes alone.

## What it does on the way back

Because the markdown holds `\"` as content, the backslash-preserving pass from #664 doubles it, and the quote-escaping pass then escapes the quote:

```
vanilla : `Betas = [\"code-execution-2025-08-25\", …
patched : `Betas = [\\\"code-execution-2025-08-25\\\", …
```

Measured inside that one prompt: `\"` **66 to 0**, `\\"` **0 to 66**. Claude Code ends up reading `\"` where the bundle originally had `"`.

The output is valid JavaScript, so `node --check`, boot and `-p` all pass. #908's parse-gate cannot see it.

## Relationship to #660

#660 reported `data-http-error-codes-reference` and `skill-stuck-slash-command` breaking JS through literal `\'` in their content, and was closed. Both of those files are still in the affected set here.

What changed is the symptom, not the cause. #664 began doubling backslashes for `"` and `'` delimiters, which turns the previously-invalid escape into valid JavaScript. The crash went away. The spurious backslash did not.

So this is the same root cause #660 was pointing at, still live, now silent instead of loud.

## Why it is separate from the `\uXXXX` issue

There is a related defect where `--apply` corrupts `\uXXXX` escapes, filed as #920. The two look identical in a diff and are not the same bug:

| | #920 | this issue |
|---|---|---|
| markdown contains | the **real character** (em dash) | the **escape text** `\"` |
| introduced by | apply, after sync was correct | sync, when writing the `.md` |
[...]

PR #923, opened 2026-07-25
Encode non-ASCII after escaping string literals, so `\uXXXX` is not double-escaped into literal text

### LINKED ISSUE #920: `--apply` silently corrupts non-ASCII in `"`/`'`-delimited system prompts: `\uXXXX` escapes get double-escaped into literal text
On a native install, `--apply` writes this into the bundle:

```
vanilla : var eam="# Claude API \u2014 C#\n\n> **Note:** The C# SDK is the official…
patched : var eam="# Claude API \\u2014 C#\n\n> **Note:** The C# SDK is the official…
```

In a double-quoted string `\\u2014` is a literal backslash followed by `u2014`. Claude Code reads the six characters `\u2014` where the em dash used to be.

The result is valid JavaScript. `node --check` passes, the binary boots, `-p` runs a turn cleanly. The parse-gate from #908 cannot see this one, because it is a content bug and not a syntax bug.

## Mechanism

Two passes disagree about what a backslash means.

1. `applySystemPrompts` detects that the bundle uses Unicode escaping and loads prompts with `shouldEscapeNonAscii = true` (`src/patches/systemPrompts.ts:146-163`). Non-ASCII in the replacement content becomes the **six-character text** `\u2014`. That is JS syntax, deliberately emitted.
2. For `"` and `'` delimiters, the backslash-preserving pass from #664 then doubles **every** backslash:

```js
if (delimiter === '"' || delimiter === "'") {
  replacementContent = replacementContent.replace(/\\/g, '\\\\');
}
```

Step 2 is right for a backslash that is literal content, which is what #664 set out to preserve. It is wrong for the backslash that step 1 just emitted as syntax, and step 1 always runs first.

Line 275 on `main` (`48ac856`), line 176 on the released v4.3.1 and v4.3.2.

## Scope

Measured on Claude Code 2.1.220 against tweakcc `48ac856`. `applySystemPrompts` has no skip-for-unchanged branch, so every prompt goes through escape+replace on every apply and an uncustomised prompt must come back byte-identical. It does not:

| | count |
|---|---|
| byte-identical after round trip | 488 / 610 |
| altered | 78 |

Bundle-wide, vanilla to patched: `\\uXXXX` (doubled, renders as literal text) **274 to 2152**. `\uXXXX` (single, renders as the real character) **71179 to 69577**.

Inside `data-claude-api-reference-csharp` alone: `\uXXXX` **38 to 0**, `\\uXXXX` **0 to 38**.

423 of the 610 generated `.md` files contain real non-ASCII, so the exposure is broad. Backtick-delimited prompts are unaffected, because #870 removed backtick from the doubling branch.

## Isolating it to apply rather than sync

The generated markdown is correct. `~/.tweakcc/system-prompts/data-claude-api-reference-csharp.md` contains 33 real em dash characters and zero occurrences of the literal text `\uXXXX`. Sync writes the character; apply turns it into escape text and then breaks the escape.

(There is a second, separate defect where sync writes *undecoded* `\"` into some `.md` files. That one is filed separately and needs a different fix; it is not this issue.)

This also does not look like #526, which reports a customised prompt not reaching Claude at all. Here the prompt does reach Claude, with a handful of characters mangled.

## Reproducing

`--apply` leaves the vanilla and patched JS side by side, so the comparison needs nothing extra:

```bash
node dist/index.mjs --apply -y

# count doubled \uXXXX (renders as literal text) in each
grep -o '\\\\u[0-9a-fA-F]\{4\}' ~/.tweakcc/native-claudejs-orig.js    | wc -l   # 274
grep -o '\\\\u[0-9a-fA-F]\{4\}' ~/.tweakcc/native-claudejs-patched.js | wc -l   # 2152
```

The second count should equal the first. Each doubled escape is a non-ASCII character that Claude Code now renders as its escape text instead.

## Suggested direction

Do not double a backslash that non-ASCII escaping just emitted. Either escape non-ASCII **after** the doubling pass, or exclude well-formed `\uXXXX` / `\xHH` sequences from the doubling.

Whatever the shape, note that `buildSearchRegexFromPieces` deliberately matches literal, `\uXXXX` and `\xHH` spellings when **searching**, so the search direction and the replace direction have different requirements. A fix that unifies them will break matching.

Worth gating on a corpus round-trip byte-identity test over all prompts, since `node --check` provably cannot detect this class.

## Environment

- tweakcc `48ac856` (also present in released v4.3.1 and v4.3.2, at line 176)
- Claude Code 2.1.220, native Linux install
- Node 24

StreamDemon:
Fixes #920.

## What

On a native install the prompt replacement is re-encoded as `\uXXXX` so it survives Bun's Latin-1 module (#853). That encoding emits JavaScript syntax, but it ran during content interpolation, before the pass that escapes the prompt's own literal backslashes (#664). So the doubling escaped the backslash the encoder had just written:

```
vanilla : var eam="# Claude API \u2014 C#\n\n> **Note:** The C# SDK is the official…
patched : var eam="# Claude API \\u2014 C#\n\n> **Note:** The C# SDK is the official…
```

`\\u2014` in a double-quoted string is a literal backslash followed by `u2014`, so Claude Code renders the escape text instead of an em dash.

This moves the encoding to the end of the delimiter-specific escaping in `applySystemPrompts`. Literal backslashes are doubled first, then non-ASCII is encoded, and the escapes the encoder produces are left intact.

## Why not just drop the doubling

#664 added it deliberately, to preserve a backslash that is part of the prompt's own text, and #660 is the report it came from. Removing it regresses both. The defect is ordering, not the doubling itself.

Backtick prompts were already unaffected, because #870 removed them from the doubling branch. This gives `"` and `'` prompts the same property.

## Shape of the change

`escapeNonAsciiChars` is exported and applied at the end of the delimiter block in `applySystemPrompts`, and no longer applied inside `applyIdentifierMapping`. `loadSystemPromptsWithRegex` loses its now-unused `escapeNonAscii` parameter; it had exactly one non-test caller.

## Verification

- Two unit tests, RED before and GREEN after: the `\uXXXX` is not double-escaped, and a prompt carrying both a literal backslash and non-ASCII gets the backslash doubled and the character encoded once, in that order. Both were mutation-checked by restoring the old ordering; both fail with exactly the reported signature.
- Full suite 460 pass / 5 skip, lint clean.
- Live end to end on a real Claude Code 2.1.220 native install, against a baseline run of `main` from an identical starting state (live binary and `native-binary.backup` both asserted clean before each run):

  | | baseline | this PR |
  |---|---|---|
  | applied / no-op / failed | 141 / 518 / 0 | 92 / 567 / 0 |
  | total entries | 659 | 659 |
  | doubled `\\uXXXX` in patched bundle | 2152 | 274 |
  | prompts round-tripping byte-identically | 488 / 610 | 533 / 610 |

  `node --check` passes, it boots, `-p` runs a turn, and the install was restored to pristine afterwards.

The applied count falls because `applied` is `oldContent !== content`: 49 prompts stop being modified at all, so they move from applied to no-op. The totals reconcile at 659 with zero failures.

## Notes

- `node --check` cannot detect this class, and neither can the parse-gate from #908. The output is valid JavaScript either way; only the text is wrong. A corpus round-trip byte-identity check is what catches it.
- This does not fix every round-trip difference. Measured with this PR alone, 33 prompts still differ, from a separate sync-side defect filed as #921.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `--apply` corrupting non-ASCII characters in single- and double-quoted system prompts on native installs.
  * Corrected escaping so Unicode `\uXXXX` sequences aren’t double-escaped or reordered when prompts contain backslashes (including em dash `\u2014`-style `\u2014`).
* **Tests**
  * Added regression coverage for the required escaping order (backslash handling before `\uXXXX` encoding), including backslash-doubling scenarios.
  * Updated interpolation tests for `<<CCVERSION>>` and `<<BUILD_TIME>>` to match the current prompt loading behavior.


xRevenantPulse:
Valid, and applied in `dc81ab4`. Single quotes do take their own branch in `applySystemPrompts` (`escapeUnescapedChar` with `'`), so the two existing cases left it unpinned.

Added a single-quoted prompt carrying both a literal backslash and an em dash, asserting the backslash is doubled and the escape is emitted once. Mutation-checked by restoring the old ordering: all three ordering tests fail, the new one included.

Suite 461 pass / 5 skip, lint clean.

PR #946, opened 2026-08-13
Recognize the extension-less 'cli' entry module of Claude Code 2.1.229+

### LINKED ISSUE #945: Claude Code 2.1.231 renames the entry module to /$bunfs/root/cli — extraction fails on every native install
## Summary

Claude Code **2.1.231** renamed its Bun entry module, and `isClaudeModule()` does not recognize the
new name. `extractClaudeJsFromNativeInstallation()` therefore returns `null` and every operation on
a native install fails with:

```
Failed to extract JavaScript from native installation: /Users/…/.local/share/claude/versions/2.1.231
```

| Claude Code | Entry module name |
| --- | --- |
| ≤ 2.1.228 | `/$bunfs/root/src/entrypoints/cli.js` |
| **2.1.231** | **`/$bunfs/root/cli`** |

`isClaudeModule()` accepts `/claude`, `claude`, `/claude.exe`, `claude.exe`,
`/src/entrypoints/cli.js`, `src/entrypoints/cli.js` — none of which match `/$bunfs/root/cli`.
Confirmed unchanged in **4.3.2** (`dist/nativeInstallation-*.mjs`, minified as `function l`).

This is the third time this exact class has broken extraction: `//root/claude` →
`//root/src/entrypoints/cli.js` (#583, #584, fixed by #585) → now `/$bunfs/root/cli`.

## Reproduction

Against a native 2.1.231 install:

```js
import { tryDetectInstallation, readContent } from 'tweakcc';

const inst = await tryDetectInstallation({ path: '/path/to/versions/2.1.231' });
await readContent(inst); // throws: Failed to extract JavaScript from native installation
```

Note that `tryDetectInstallation` still succeeds and reports `version: '2.1.231', kind: 'native'` —
only the module lookup fails.

Dumping the module table of 2.1.231 (Mach-O `__BUN`/`__bun`, u64 section header, 15 modules,
52-byte module structs, `entryPointId: 0`):

```
 0 "/$bunfs/root/cli"                          contents=25410680   <-- the bundle
 1 "/$bunfs/root/image-processor.js"           contents=2171
 2 "/$bunfs/root/audio-capture.js"             contents=2169
 …
13 "/$bunfs/root/payload.template.html.asset"  contents=2229365
14 "/$bunfs/root/url-handler.node"             contents=336864
```

For comparison, 2.1.228 module 0 is `/$bunfs/root/src/entrypoints/cli.js`.

## Secondary hazard: the repack path fails silently

`rebuildBunDataBlob` selects the module to overwrite with the *same* predicate:

```js
u = n && l(o) ? n : s(e, t.contents);   // new contents only if the name matches
```

When nothing matches, every module keeps its original contents and the binary is repacked
**unchanged**. Inside tweakcc's own flow this is unreachable because `readContent` fails first, but a
caller that supplies its own source and calls `repackNativeInstallation()` directly gets a
successfully-written, silently unpatched binary instead of an error. Worth failing loudly there
regardless of how the name lookup is fixed.

## Suggested fix

The blob already records which module is the entry point — `entryPointId` is parsed into
`bunOffsets` and is `0` here. Selecting by that instead of by name is immune to renames:

```js
// in extractClaudeJsFromNativeInstallation / rebuildBunDataBlob
const isTarget = (name, index) => index === bunOffsets.entryPointId || isClaudeModule(name);
```

Keeping `isClaudeModule` as a fallback preserves current behavior for any layout where the entry
point is not the bundle. Simply adding `/$bunfs/root/cli` to the list would unblock 2.1.231, but the
next rename breaks it again.

## Note for anyone testing a fix

The entry module also carries ~190 MB of Bun bytecode (`// @bun @bytecode @bun-cjs`), which the
repack preserves verbatim. That bytecode does **not** take precedence over the replaced source — I
injected a canary into the extracted JS of 2.1.231, repacked, and it printed at startup. So patching
works normally once the module can be found; the bytecode is not an additional obstacle.

## Environment

- tweakcc 4.3.0 and 4.3.2 (identical predicate in both)
- Claude Code 2.1.231, native install
- macOS 15 / arm64 (Darwin 25.5.0), Node 24.14.1

Verified on Mach-O only. The predicate is format-independent, so ELF and PE installs of 2.1.231
should be affected identically, but I have not run them.

ahalekelly:
🤖 Generated by [Claude Code](https://claude.com/claude-code)

Closes #945

Claude Code 2.1.229 renamed the Bun compile entry module from `src/entrypoints/cli.js` to extension-less `cli` — `/$bunfs/root/cli` on POSIX, `B:/~BUN/root/cli` on Windows. `isClaudeModule()` no longer matches it, so both `unpack` and `repack` fail on 2.1.229 native binaries with `Error: Could not extract JS from native binary`. Same failure mode as #584 when 2.1.69 last renamed the module.

The rename looks deliberate and permanent: the Claude Agent SDK's `extractFromBunfs.js` extracts the embedded CLI to a temp file named after the module's basename, and comments that a native binary must not come out named `cli.js` because `ProcessTransport.isNativeBinary()` picks the spawn strategy by extension.

## Testing

- `unpack` on the published 2.1.229 binaries for darwin-arm64, linux-x64, and win32-x64 (from the `@anthropic-ai/claude-code-*` npm packages): all extract successfully (~25.4 MB of JS each); all fail without this change.
- `unpack` on 2.1.228 darwin-arm64 still works (old `src/entrypoints/cli.js` name, no regression).
- Round-trip: `repack` of the extracted JS into a copy of the 2.1.229 darwin-arm64 binary succeeds and the repacked binary runs (`--version` → `2.1.229 (Claude Code)`).
- `npm run lint` and the vitest suite (510 tests) pass. Pre-commit hook steps were run manually — the hook invokes pnpm, which this machine doesn't have.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recognition of Claude Code command-line modules, including modules named `cli` or ending in `/cli`.

### DIFF
--- a/src/nativeInstallation.ts
+++ b/src/nativeInstallation.ts
@@ -227,7 +227,11 @@ function isClaudeModule(moduleName: string): boolean {
     moduleName.endsWith('/claude.exe') ||
     moduleName === 'claude.exe' ||
     moduleName.endsWith('/src/entrypoints/cli.js') ||
-    moduleName === 'src/entrypoints/cli.js'
+    moduleName === 'src/entrypoints/cli.js' ||
+    // Claude Code >= 2.1.229 names the entry module `cli` with no extension
+    // (`/$bunfs/root/cli` on POSIX, `B:/~BUN/root/cli` on Windows).
+    moduleName.endsWith('/cli') ||
+    moduleName === 'cli'
   );
 }
 


bl-ue (approved):
Thank you!

